PR #2, opened 2024-04-03
Port pandas-ts code

hombit:
This ports code from https://github.com/lincc-frameworks/pandas-ts


dougbrn (approved):
I've read through this a few times, I don't fully grasp every detail yet, but nothing stood out to me as obviously problematic. Just had one nit on a test name

PR #5, opened 2024-04-03
initialize nestedframe

dougbrn:
Initializes the NestedFrame module. Adds a few properties and an initial add_nested function.

### DIFF
--- a/src/nested_pandas/__init__.py
+++ b/src/nested_pandas/__init__.py
@@ -1,4 +1,5 @@
 from .example_module import greetings, meaning
+from .nestedframe import NestedFrame  # noqa
 
 # Import for registering
 from .series.accessor import NestSeriesAccessor  # noqa: F401
--- a/src/nested_pandas/__init__.py
+++ b/src/nested_pandas/__init__.py
@@ -1,8 +1,8 @@
 from .example_module import greetings, meaning
-from .nestedframe import NestedFrame  # noqa
+from .nestedframe import NestedFrame
 
 # Import for registering
 from .series.accessor import NestSeriesAccessor  # noqa: F401
 from .series.dtype import NestedDtype
 
-__all__ = ["greetings", "meaning", "NestedDtype"]
+__all__ = ["greetings", "meaning", "NestedDtype", "NestedFrame"]
--- a/src/nested_pandas/nestedframe/__init__.py
+++ b/src/nested_pandas/nestedframe/__init__.py
@@ -1,0 +1,1 @@
+from .core import *  # noqa
--- a/src/nested_pandas/nestedframe/__init__.py
+++ b/src/nested_pandas/nestedframe/__init__.py
@@ -1,1 +1,1 @@
-from .core import *  # noqa
+from .core import NestedFrame  # noqa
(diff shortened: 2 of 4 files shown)


src/nested_pandas/__init__.py
```diff
@@ -1,4 +1,5 @@
 from .example_module import greetings, meaning
+from .nestedframe import NestedFrame  # noqa
```
hombit, line 2: Let's remove `noqa` and add to `__all__`?

src/nested_pandas/nestedframe/__init__.py
```diff
@@ -0,0 +1 @@
+from .core import *  # noqa
```
hombit, line 1: Do we really need star-import here? It would add `pd` and other stuff we don't really need

src/nested_pandas/nestedframe/core.py
```diff
@@ -0,0 +1,61 @@
...
+    @property
+    def _constructor_expanddim(self) -> Self:  # type: ignore[name-defined] # noqa: F821
+        return NestedFrame
+
+    @property
+    def all_columns(self) -> dict:
+        """returns a dictionary of columns for each base/nested dataframe"""
+        all_columns = {"base": self.columns}
+        for column in self.columns:
+            if hasattr(self[column], "nest"):
```
hombit, line 31: I would check column's dtype instead, it should be `NestedDtype`. No code suggestion here, because we need to import it first
```python
from nested_pandas import NestedDtype

if isinstance(self[column].dtype, NestedDtype):
```

src/nested_pandas/nestedframe/core.py
```diff
@@ -0,0 +1,61 @@
...
+                nest_cols = self[column].iloc[0].columns  # TODO: Improve access to columns
```
hombit, line 32:
```suggestion
                nest_cols = self[column].nest.fields()
```

src/nested_pandas/nestedframe/core.py
```diff
@@ -0,0 +1,61 @@
...
+                all_columns[column] = nest_cols
+        return all_columns
+
+    @property
+    def nested_columns(self) -> list:
+        """retrieves the base column names for all nested dataframes"""
+        nest_cols = []
+        for column in self.columns:
+            if hasattr(self[column], "nest"):
```
hombit, line 41:
```suggestion
            if isinstance(self[column].dtype, NestedDtype):
```

src/nested_pandas/nestedframe/core.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def _is_known_hierarchical_column(self, colname) -> bool:
+        """Determine whether a string is a known hierarchical column name"""
+        if "." in colname:
+            left, right = colname.split(".")
+            if left in self.nested_columns:
+                return right in self.all_columns[left]
+            else:
+                return False
+        else:
```
hombit, line 53: It is a matter of taste, but I believe we don't need these `else:`s here

hombit (commented):
Looks great! Please see some suggestions

dougbrn:
Thanks for the comments, I've addressed them all in the latest commit

hombit (approved):
Looks great

Issue #10, opened 2024-04-04
MVP: Wrap Query Function - allow nested access


dougbrn:
**Feature request**
Blocks #9. Implement NestedFrame.query to allow hierarchical column access, e.g.
```
NestedFrame.query("my_nested.a > 2")
```

The current thinking is that this returns back the full NestedFrame base layer, but would update the nested dataframe "my_nested" with the result of the query "a>2".

This can be accomplished by wrapping query and doing some string evaluation first to determine which layer to send the query to. I think to start we should limit a query string to only target one layer.

**Before submitting**
Please check the following:

- [x] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [x] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [x] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

PR #19, opened 2024-04-05
Initial query wrapping

dougbrn:
## Change Description

Addresses #10.

## Solution Description

This PR wraps the pandas query function. Before executing any query, it first checks the input query string for spacing between the various components. I needed this to essentially be able to split each component apart and check to see if it was targeting a nested column using the "nested_frame.nested_column" access pattern. It then builds a set of subqueries for each targeted layer. At the moment, I limited this to only having one layer be targetable at a time, but we could allow multiple queries to execute within the query call down the road.

One note, is that using the "." is slightly overloaded in Pandas eval syntax. You can use "column.property" within eval normally. We override this by splitting on it and removing the dot, so some subset of queries full functionality is probably limited. I think this implementation should be sufficient for the MVP however, and we can come back to make this more robust in a later stage of the project.

## Code Quality
- [ ] My code contains relevant comments and necessary documentation

### Bug Fix Checklist
- [ ] My fix includes a new test that breaks as a result of the bug (if possible)

### New Feature Checklist

### DIFF
--- a/src/nested_pandas/nestedframe/utils.py
+++ b/src/nested_pandas/nestedframe/utils.py
@@ -1,0 +1,34 @@
+def _ensure_spacing(expr) -> str:
+    """Ensure that an eval string has spacing"""
+    single_val_operators = ["+", "-", "*", "/", "%", ">", "<", "|", "&", "~", "="]  # omit "(" and ")"
+    check_for_doubles = ["=", "/", "*", ">", "<"]
+    double_val_operators = ["==", "//", "**", ">=", "<="]
+    expr_list = [*expr]
+    i = 0
+    spaced_expr = ""
+    while i < len(expr_list):
+        if expr_list[i] not in single_val_operators:
+            spaced_expr += expr_list[i]
+        else:
+            if expr_list[i] in check_for_doubles:
+                if "".join(expr_list[i : i + 2]) in double_val_operators:
+                    if spaced_expr[-1] != " ":
+                        spaced_expr += " "
+                    spaced_expr += "".join(expr_list[i : i + 2])
+                    if expr_list[i + 2] != " ":
+                        spaced_expr += " "
+                    i += 1  # skip ahead an extra time
+                else:
+                    if spaced_expr[-1] != " ":
+                        spaced_expr += " "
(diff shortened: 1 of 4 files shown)


src/nested_pandas/nestedframe/utils.py
```diff
@@ -0,0 +1,34 @@
+def _ensure_spacing(expr) -> str:
+    """Ensure that an eval string has spacing"""
+    single_val_operators = ["+", "-", "*", "/", "%", ">", "<", "|", "&", "~", "="]  # omit "(" and ")"
+    check_for_doubles = ["=", "/", "*", ">", "<"]
+    double_val_operators = ["==", "//", "**", ">=", "<="]
```
hombit, line 5: I would suggest using sets for better semantics 
```suggestion
    single_val_operators = {"+", "-", "*", "/", "%", ">", "<", "|", "&", "~", "="}  # omit "(" and ")"
    check_for_doubles = {"=", "/", "*", ">", "<"}
    double_val_operators = {"==", "//", "**", ">=", "<="}
```

src/nested_pandas/nestedframe/utils.py
```diff
@@ -0,0 +1,34 @@
...
+    expr_list = [*expr]
```
hombit, line 6: I'm a bit confused by this line. Why do we do `[*expr]` instead of `list(expr)`? I believe we don't even need it to be a list: `expr_list[i]` is a single character string, `expr_list[i : i + 2]` is a slice over a string, which is also a string.

hombit (approved):
Looks great, thank you!

src/nested_pandas/nestedframe/utils.py
dougbrn: Good call, implemented in latest commit

src/nested_pandas/nestedframe/utils.py
dougbrn: I'm not sure why I opted to split the string out into a list, I was trying a few different solution methods and so perhaps this just stuck as I was working through it. You're correct in that we can just operate on this as a string, I've updated it to do so in the latest commit

PR #27, opened 2024-04-12
Propagate index name in `to_flat`

wilsonbb:
## Change Description
Currently when we use `to_flat` on a packed dataframe the name of its original index (which still exists in its pandas series form) is lost. See the following notebook screenshot

Screenshot 2024-04-12 at 9 45 36 AM

This can be fixed by simply having `to_flat` reference the previous index name when creating its result dataframe. 

## Code Quality

### Bug Fix Checklist
- [x] My fix includes a new test that breaks as a result of the bug (if possible)


src/nested_pandas/series/accessor.py
```diff
@@ -99,7 +99,8 @@ def to_flat(self, fields: list[str] | None = None) -> pd.DataFrame:
                 name=field,
                 copy=False,
             )
-        return pd.DataFrame(flat_series)
+        # Note that index should now not be none since we iterated over at least one field
+        return pd.DataFrame(flat_series, index=pd.Series(index, name=self._series.index.name))
```
hombit, line 103: Could we put it to line 95 instead?

hombit (changes requested):
Thank you! TIL about index names

src/nested_pandas/series/accessor.py
hombit: I mean I believe that it would be enough to put the index into each pd.Series instead of putting it into result dataframe

wilsonbb: Yeah, that's reasonable. Done.

hombit (approved):

Issue #14, opened 2024-04-04
MVP: Implement a batch/apply style function


dougbrn:
**Feature request**
Blocks #9. We want to apply a function, such as a periodogram function, to all lightcurves stored within a nested structure. We should implement a simple function for applying this function to the lightcurves. With TAPE, this motivated the batch function for grouping the sources and applying a function to them. Our data structures motivate a custom function a bit less, as the data is pre-grouped. So using the base apply may potentially be fine, however we may be able to provide a more intuitive API function for doing this operation and can potentially optimize for how it accesses the lightcurve data under the hood. Usage should look something like this:

```
mean_flux = nf.batch(my_mean_flux, "my_nested.flux")
```

**Before submitting**
Please check the following:

- [x] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [x] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [x] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

wilsonbb:
A small suggestion here is that I would avoid the name "batch" here since it has the connotation that we're doing batch processing of splitting the job into multiple computable batches. That connotation makes less sense here where there is no applied parallelism.

PR #32, opened 2024-04-18
Add NestedFrame.reduce

wilsonbb:
## Change Description
To address https://github.com/lincc-frameworks/nested-pandas/issues/14, here we provide an easy way for users to apply a custom function to a `NestedFrame`, where the user-provided function is applied to each row of the base layer of the `NestedFrame` along the specified columns (including the ability to specify nested columns packed into that row). The result of the function is then another nested frame which can be worked with separately or joined back to the original frame. 

We opted for the provisional name `reduce` since it provides to the user function "base" layer columns as scalars and hierarchical columns as collections allowing for the "reduction" of the hierarchical data to a frame matching the indices of the base layer. For other cases of providing a custom function, the user can still. the use the `apply` API.

## Solution Description
This solution wraps around [pandas.DataFrame.apply](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.apply.html) to apply the user function to each row of the nested frame. Note that for columns in the "nested" frames, these are already packed as a dataframe within their respective cells of that row. So we do not need to unpack and flatten in the case where a user requests hierarchical columns.

## Code Quality

### New Feature Checklist
- [x] I have added unit/End-to-End (E2E) test cases to cover my new feature


src/nested_pandas/nestedframe/core.py
```diff
@@ -310,3 +315,85 @@ def dropna(
             )
         )
         return new_df
+
+    def reduce(self, func, *args, **kwargs)  -> NestedFrame:
```
dougbrn: I'm going back and forth on this name. I like that it matches the idea of getting a lower dimensional result from a potential higher-dimensionality dataset, but am struggling with the idea that nothing about it suggests that this is where you would throw your analysis code at the package... Can always just hammer this in the docs though

src/nested_pandas/nestedframe/core.py
```diff
@@ -310,3 +316,85 @@ def dropna(
...
+        def translate_cols(frame, layer, col):
+            if layer == "base":
+                # We pass the "base" column as a scalar
+                return frame[col]
+            return frame[layer][col].to_numpy()
+
+        # Note that this applys the function to each row of the nested dataframe. For
+        # the columns within packed frames, note taht we're directly accessing the dataframe
+        # within the cell of that row without having to unpack and flatten.
+        result = self.apply(
```
dougbrn, line 390: Apply seems correct here, but there's a potential for it being slow. I will be doing some benchmarking on this once it goes in and may suggest an alternative in the near future.

src/nested_pandas/nestedframe/core.py
```diff
@@ -310,3 +316,85 @@ def dropna(
...
+            Keyword arguments to pass to the function.
+
+        Returns
+        -------
+        `NestedFrame`
+            `NestedFrame` with the results of the function applied to the columns of the frame.
+
+        Notes
+        -----
+        The recommend return value of func should be a `pd.Series` where the indices are the names of the
```
dougbrn, line 344: pd.Series is fine for multi-column results, but for functions that return just a single value pd.Series will be unnecessary overhead (as it has to generate a series for each result). We can either add a note here, or this can be something to remember for documentation (particularly at the dask level where we really care about performance).

tests/nested_pandas/nestedframe/test_nestedframe.py
```diff
@@ -222,3 +222,83 @@ def test_dropna_errors():
...
+    expected_max_c = [4, 10, 4]
+    expected_max_d = [7, 5, 9]
+    expected_max_e = [9, 23, 4]
+
+    # Test that we raise an error when no arguments are provided
+    with pytest.raises(ValueError):
+        nf.reduce(get_max)
+
+    # Batch only on columns in the first packed layer
+    result = nf.reduce(get_max, "packed.c", "packed.d")
```
dougbrn, line 270: Awesome.

dougbrn (approved):
Looks good! I don't have any technical blockers, just a couple of un-actionable thoughts.

src/nested_pandas/nestedframe/core.py
wilsonbb: Yeah as discussed offline we should definitely look more into it.

src/nested_pandas/nestedframe/core.py
wilsonbb: Yeah very good points. I don't have strong attachments so can always hammer out later.

Issue #33, opened 2024-04-22
`to_lists()` with column subset much slower than all columns when chained to `apply(..., axis=1)`


dougbrn:
**Bug report**

The following code is not a reproducible example because the data creation is not included, but applying this to any nestedframe should yield similar timing results.

```
nf.ztf_sources.nest.to_lists(["mjd","flux"]).apply(lambda x: x, axis=1) # 461ms
nf.ztf_sources.nest.to_lists().apply(lambda x: x, axis=1) # 34.4 ms
```

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.
- [ ] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

hombit:
I believe it may be unrelated to our code and it is (again) a pandas dtypes issue.

First, let me re-write the code with built-in generator:
```python
from nested_pandas.datasets import generate_data

ndf = generate_data(n_base=1000, n_layer=100)
%timeit ndf.nested.nest.to_lists().apply(lambda x: x, axis=1)  # 8.2ms
%timeit ndf.nested.nest.to_lists(["t", "flux"]).apply(lambda x: x, axis=1)  # 76ms
%timeit ndf.nested.nest.to_lists(["t", "band"]).apply(lambda x: x, axis=1)  # 8.1ms
```

The difference between second line and two other lines is in dtypes of the columns: "t" and "flux" are float64 columns, while "band" is a string column. In both cases `pandas` creates a `pd.Series` from each row. When dtypes are different, pd.Series has `object` dtype, and when they are the same, well, `pandas` casts pd.Series to this dtype. Surprisedly, it turns to be a quiet expensive operation:

```python
import numpy as np
import pandas as pd
import pyarrow as pa

zeros = pa.scalar(np.zeros(100, dtype=float))
%timeit pd.Series([zeros, zeros])  # 13.8 μs
%timeit pd.Series([zeros, zeros], dtype=pd.ArrowDtype(zeros.type))  # 160 μs
```

What we can do it calling [`.assign` with `raw=True`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html), which is a close equivalent of our `reduce()`:
```python
[...]

PR #34, opened 2024-04-22
speedup with to_lists()

dougbrn:
This PR uses `to_lists()` in place of `to_flat()` and `apply()` for the `by=None` `count_nested` case and in `reduce` to offer a 5-10x speedup.

It does introduce one limitation to `reduce` in that the nested structures cannot have overlapping column names, this will be less of an issue once #33 is addressed. I propose we wait to resolve this overlap issue until after #33 and just make a separate issue for now.

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -387,11 +387,11 @@ class NestedFrame(pd.DataFrame):
         for layer in layers:
             if layer == "base":
                 columns = [col[1] for col in requested_columns if col[0] == layer]
-                apply_df = apply_df.join(self[columns], how="right")
+                apply_df = apply_df.join(self[columns], how="outer")
             else:
                 # TODO: It should be faster to pass these columns to to_lists, but its 20x slower
                 # columns = [col[1] for col in requested_columns if col[0] == layer]
-                apply_df = apply_df.join(self[layer].nest.to_lists(), how="right")
+                apply_df = apply_df.join(self[layer].nest.to_lists(), how="outer")
 
         # Translates the requested columns into the scalars or arrays we pass to func.
         def translate_cols(frame, layer, col):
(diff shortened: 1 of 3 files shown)


src/nested_pandas/nestedframe/core.py
```diff
@@ -379,22 +379,32 @@ def my_sum(col1, col2):
...
+        layers = np.unique([col[0] for col in requested_columns])
+
+        # build a flat dataframe with array columns to apply to the function
+        apply_df = NestedFrame()
+        for layer in layers:
+            if layer == "base":
+                columns = [col[1] for col in requested_columns if col[0] == layer]
+                apply_df = apply_df.join(self[columns], how="right")
+            else:
+                # TODO: It should be faster to pass these columns to to_lists, but its 20x slower
```
wilsonbb, line 392: Huh this is interesting

wilsonbb (approved):
An exciting speedup! Just added a comment suggesting we use a different form of join.

src/nested_pandas/nestedframe/core.py
```diff
@@ -379,22 +379,32 @@ def my_sum(col1, col2):
...
 
+        # find targeted layers
+        layers = np.unique([col[0] for col in requested_columns])
+
+        # build a flat dataframe with array columns to apply to the function
+        apply_df = NestedFrame()
+        for layer in layers:
+            if layer == "base":
+                columns = [col[1] for col in requested_columns if col[0] == layer]
+                apply_df = apply_df.join(self[columns], how="right")
```
wilsonbb, line 390: I think we should use an "inner" join here to ensure we a) always have a result index that matches the base layer index and b) the order in whi
ch we join doesn't matter.

For example, with `reduce(func, "packed2.e", "packed.c")` on the following:

Screenshot 2024-04-23 at 10 35 59 AM

depending on the order of the join, we'll either have a result with no values for index 0 or no values for index 1.  And in this specific cases because `np.unique` returns the unique values in sorted order, we apply the join in alphabetical order of the base layers (so `join(packed, packed2, how="right")` rather than the order in which the columns were specified).

wilsonbb: Correction, I edited the above to be "outer" rather than "inner" join

PR #58, opened 2024-05-07
add a few benchmarks

### LINKED ISSUE #37: Define a few ASV benchmarks
We should define a small set of benchmarks up front. Particularly important to benchmark:

* data loading
* filtering
* reduce
* low-level extension array

dougbrn:
Resolves #37. Adds a few more benchmarks. I opted to just test add_nested instead of adding parquet files for now, as in principle our package inherits the performance of parquet reading from pandas. It may still be good to add in the future (or now if others feel strongly) to check for any regressions, but it would be slightly redundant with whatever testing Pandas itself already does.

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -6,7 +6,7 @@ https://asv.readthedocs.io/en/stable/writing_benchmarks.html."""
 import numpy as np
 import pandas as pd
 import pyarrow as pa
-from nested_pandas import NestedDtype
+from nested_pandas import NestedDtype, NestedFrame, datasets
 
 
 class AssignSingleDfToNestedSeries:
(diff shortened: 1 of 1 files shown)


benchmarks/benchmarks.py
```diff
@@ -98,3 +98,93 @@ def time_run(self):
...
+        """Set up the benchmark environment"""
+        self.nf = datasets.generate_data(self.n_base, self.n_nested)
+
+    def run(self):
+        """Run the benchmark."""
+        # Apply base layer query
+        self.nf = self.nf.query("a > 0.2")
+
+        # Apply nested layer query
+        self.nf = self.nf.query("nested.band == 'g'")
```
hombit, line 182: Would we like to split it into two benchmarks?

hombit (approved):
Looks great

benchmarks/benchmarks.py
dougbrn: I think it's fine, as having a benchmark for the top layer query is just timing pandas code.

dougbrn: on second thought, for consistency with above I opted to just remove the top layer query.

Issue #87, opened 2024-05-22
.nest operations do not update series dtype


hombit:
**Bug report**

`.nest` operations changing `dtype` of the underlying `NestedExtensionArray` do not actually change `series.dtype`. These operations include field addition and deletion, assignment of a field with different dtype, etc. It is a problem, because it creates an inconsistent series with data and dtype to have different fields.

```python
from nested_pandas.datasets import generate_data

df = generate_data(5, 3)
# Passes
assert df['nested'].dtype == df['nested'].array.dtype
del df['nested'].nest['band']
# Raises 
assert df['nested'].dtype == df['nested'].array.dtype
```

I don't know the exact technical reason for this behavior, it is not obvious from the `pandas` code how and where `dtype` is cached - it looks like it *should* be just the same as the underlying `array`'s dtype.

I think that the initial design decision of making `.nest` mutating structure of the underlying array was wrong: while it is fine to change values, `pandas` doesn't imply dtype changes to be done through accessors. We should rework the accessor to return new `pd.Series` when changing field structure or types, instead of doing it "in-place".

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.
- [x] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

PR #91, opened 2024-05-29
Fixes for accessor mapping methods

### LINKED ISSUE #62: NestedAccessor's .pop and .pop_field() do the same
**Feature request**

It is confusing, we should remove `.pop_field`

**Before submitting**
Please check the following:

- [ ] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.

hombit:
This PR addresses issue #87 and removes most of the in-place modification methods from the `NestedAccessor`. The problem with field modification was that `pd.Series.dtype` would not update after changing the underlying `NestedAccessor.dtype`. The only modification we can still perform through the accessor interface is changing the field (subcolumn) data without altering its dtype.

Note: When I mention "in-place" here, I don't mean that we do not copy data — we do copy the arrays we modify. However, this process is transparent from the user's and `pd.Series`'s point of view. Therefore, we are effectively changing `pd.Series` in-place, while the underlying memory blobs remain immutable.

It also fixes #62, because we wouldn't have both methods anymore.

### DIFF
--- a/src/nested_pandas/series/ext_array.py
+++ b/src/nested_pandas/series/ext_array.py
@@ -35,7 +35,7 @@
 # typing.Self and "|" union syntax don't exist in Python 3.9
 from __future__ import annotations
 
-from collections.abc import Iterator, Sequence
+from collections.abc import Iterable, Iterator, Sequence
 from typing import Any, Callable, cast
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


src/nested_pandas/series/ext_array.py
```diff
@@ -676,43 +676,94 @@ def view_fields(self, fields: str | list[str]) -> Self:  # type: ignore[name-def
...
             The name of the field.
         value : ArrayLike
             The 'flat' array of values to be set.
+        keep_dtype : bool, default False
+            Whether to keep the original dtype of the field. If True,
+            now new field will be created, and the dtype of the existing
+            field will be kept. If False, the dtype of the field will be
+            inferred from the input value.
         """
         # TODO: optimize for the case when the input is a pa.ChunkedArray
```
dougbrn, line 697: Does this PR address the TODO here at all?

dougbrn (approved):
Looks great!

src/nested_pandas/series/ext_array.py
hombit: Not yet =(

PR #94, opened 2024-05-29
Rename ext_array._pa_array to _chunked_array

hombit:
`pandas`' `ArrowExtensionArray` uses `._pa_array` for the underlying array data, but it is a bit confusing because it is actually a chunked array. This PR changes this name in our implementation and also introduces a public `chunked_array` property.

It adds a data model description to `NestedExtensionArray` docs.


dougbrn (approved):
Nice, looks good

PR #96, opened 2024-05-30
Reimplementation of NestedFrame.reduce()

hombit:
Here is a re-implementation of reduce without using `df.join` and `df.assign`. It is shorter and runs faster


wilsonbb (approved):
This looks cool, thanks for this!

PR #103, opened 2024-06-05
Custom HTML Rendering for Nested Columns

### LINKED ISSUE #41: Add a custom html dataframe rendering view of nested columns
We have control over the dataframe output within Pandas: https://pandas.pydata.org/docs/user_guide/style.html

When left unstyled, Pandas tries to print the dataframe directly into the cell values. We should have a custom rendering that changes this to be some simple representative value, like: `<NestedFrame>`. Optionally we could try to give this some more descriptive metadata, but I worry about the overhead of calculating such information if the user doesn't need it. e.g. `<NestedFrame (length=350)>`

dougbrn:
## Change Description

Resolves #41 .

Also, I discovered some instability with the SDSS spectra notebook when updating this PR that's completely unrelated. I believe I have resolved it with a call to clear the download cache, and put that fix directly into this PR.

## Solution Description

Update (Feb 12, 2025): This PR now contains a more functional representation for nested data:

Screenshot 2025-02-12 at 1 11 52 PM

Nestedframes now provide a preview row of the nested data alongside the dimensionality of the nested dataframes. In addition the nested columns are represented in the first row. Alignment is a bit lacking, but this seems to be a big improvement over what is present currently. This is only for HTML output (e.g. jupyter)


dougbrn:
I've revisted this briefly, this is what this PR can implement. I'm not happy with the loss of information as described above, and am beginning to believe the pandas styler object just does not give us the tools to do something about this. 
Screen Shot 2024-07-11 at 4 38 33 PM

src/nested_pandas/nestedframe/core.py
```diff
@@ -274,6 +274,52 @@ def nested_columns(self) -> list:
...
+        repr = repr.format(
+            {col: repack_row for col in self.nested_columns}, subset=pd.IndexSlice[self.index[1] :]
+        )
+
+        # Recover some truncation formatting, limited to head truncation
+        # Use half of pandas sizes to account for large row height
+        if repr.data.shape[0] > pd.get_option("display.max_rows") // 2:
+            html_repr = repr.to_html(max_rows=pd.get_option("display.min_rows") // 2)
+        else:
+            html_repr = repr.to_html(max_rows=0)
```
gitosaurus, line 316: Ok, it's intentional here, I see. So this means "no truncation" when the number of rows is greater than half of `display.max_rows`; correct? (In practice this looks like it also includes row elision, in my sandbox.) Or do you mean to clip it to `1` row or otherwise be minimal?

gitosaurus (approved):
Looks good, but I wonder about one edge case, in my comments. Maybe add a code comment about rationale.

src/nested_pandas/nestedframe/core.py
```diff
@@ -274,6 +274,52 @@ def nested_columns(self) -> list:
...
+            {col: repack_first_cell for col in self.nested_columns}, subset=self.index[0]
+        )
+        repr = repr.format(
+            {col: repack_row for col in self.nested_columns}, subset=pd.IndexSlice[self.index[1] :]
+        )
+
+        # Recover some truncation formatting, limited to head truncation
+        # Use half of pandas sizes to account for large row height
+        if repr.data.shape[0] > pd.get_option("display.max_rows") // 2:
+            html_repr = repr.to_html(max_rows=pd.get_option("display.min_rows") // 2)
```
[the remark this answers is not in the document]
dougbrn: You're right in that you get zeros here, and the behavior of 0 is to display everything which is not quite intended. I think this has a larger issue, in that the numbers that a user might set don't map to what they see in the nestedframe. I removed the `// 2` here in favor of just changing the display defaults for nestedframes so that a user can change them accurately on their end if they wish.

src/nested_pandas/nestedframe/core.py
dougbrn: Right, effectively this is just ensuring that all rows are represented if under the `max_rows` threshold, and otherwise a truncated view with `min_rows` is provided. Again removing the `// 2` as per above. I added a comment to explain this more.

PR #107, opened 2024-06-10
Reduce Docstring: Update with note on using dictionary outputs

### LINKED ISSUE #106: Update `reduce` function recommendation
Following https://github.com/lincc-frameworks/nested-dask/issues/27, we should make sure to recommend users return dictionaries in their custom functions. We currently recommend a pd.Series which will be significantly slower and should not be encouraged behavior.

dougbrn:
resolves #106. Partially addresses https://github.com/lincc-frameworks/nested-dask/issues/27 as well.


hombit (approved):
Looks great, thank you

Issue #120, opened 2024-07-08
Refactor add_nested to nested_join


hombit:
Currently, the only supported way to add nested columns is by packing them by index and assigning them to NestedFrame with the same index values. However, as was found during the LSDB integration (https://github.com/astronomy-commons/lsdb/pull/368, https://github.com/astronomy-commons/lsdb/issues/367), it is important to pack by the value column instead of index.

Additionally, it would be nice to make the interface closer to pandas’ join by supporting `on` and `how` arguments.

The following steps are proposed:
- [ ] Rename `add_nested` to `nested_join`
- [ ] https://github.com/lincc-frameworks/nested-pandas/issues/115
- [ ] https://github.com/lincc-frameworks/nested-pandas/issues/117
- [ ] https://github.com/lincc-frameworks/nested-pandas/issues/116

hombit:
Everything but renaming is implemented in v0.3.0

PR #125, opened 2024-07-25
Add "how" argument to add_nested

wilsonbb:
## Change Description

Adds a `how` argument in `add_nested` to align with a traditional pandas join and in preparation from the change from `add_nested` to `join_nested` (see https://github.com/lincc-frameworks/nested-pandas/issues/120) 

## Solution Description

The previous `add_nested` implementation packed the nested frame into a single column which is then assigned to our base layer. Here we simply add this column using a traditional join and then make use of the join's `how` argument. Since our previous implementation produces the same results as the "left" argument, I was concerned that replacing the single column assignment with a join might decrease performance. However a preliminary benchmarking shows no notable change. 

Screenshot 2024-07-25 at 1 17 40 PM
 
Also of note here is that for the case of `how="right"`, the nature of the packer still means that the result is implicitly using the index of our base layer but now drops values from it that were not present in the index of our "right" nested layer.    

## Code Quality

### New Feature Checklist

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -66,7 +66,12 @@ class NestedFrame(pd.DataFrame):
         return colname in self.columns or self._is_known_hierarchical_column(colname)
 
     def add_nested(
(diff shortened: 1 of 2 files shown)


dougbrn (approved):
Nice looks good!

Issue #59, opened 2024-05-07
Nested-pandas query does not handle scientific notation


dougbrn:
**Bug report**
The following code block fails in nested_pandas, but succeeds in pandas.

```
from nested_pandas import NestedFrame
base = NestedFrame({"a":[1,2,3]}, index=[0,1,2])
base.query("a > 1e-1")
```
The query parsing code most likely tries to read the "e" as a column name and splits it apart, we should add additional handling for this.

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.
- [x] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

Issue #146, opened 2024-09-20
query("nested.b.isna()") fails


hombit:
**Bug report**

We cannot use built-in methods in queries because of an issue in the query parsing code:

```python
from nested_pandas import NestedFrame

nf = NestedFrame.from_flat(NestedFrame({'a': [1, 2], 'b': [3, None]}, index=[1,1]), base_columns=[])
nf.query("nested.b.isna()")
```

```
File ~/.virtualenvs/v/lib/python3.12/site-packages/nested_pandas/nestedframe/core.py:357, in NestedFrame.query(self, expr)
    313 """
    314 Query the columns of a NestedFrame with a boolean expression. Specified
    315 queries can target nested columns in addition to the typical column set
   (...)
    353 >>> df.query("mynested.a > 2")
    354 """
    356 # Rebuild queries for each specified nested/base layer
--> 357 exprs_to_use = self._split_query(expr)
    359 # For now (simplicity), limit query to only operating on one layer
    360 if len(exprs_to_use.keys()) != 1:

File ~/.virtualenvs/v/lib/python3.12/site-packages/nested_pandas/nestedframe/core.py:293, in NestedFrame._split_query(self, expr)
    291 expr_slice = split_expr[i].strip("()")
    292 # Check if it's a nested column
--> 293 if self._is_known_hierarchical_column(expr_slice):
    294     nested, colname = split_expr[i].split(".")
    295     current_focus = nested.strip("()")

File ~/.virtualenvs/v/lib/python3.12/site-packages/nested_pandas/nestedframe/core.py:58, in NestedFrame._is_known_hierarchical_column(self, colname)
     56 """Determine whether a string is a known hierarchical column name"""
     57 if "." in colname:
---> 58     left, right = colname.split(".")
     59     if left in self.nested_columns:
     60         return right in self.all_columns[left]

ValueError: too many values to unpack (expected 2)
```

I believe it is related to #144, but probably would need a different fix

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.
- [ ] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

Issue #144, opened 2024-09-17
Column name parsing fail for multiple-dot names


hombit:
**Bug report**

I have a column name "R.A." and `lsdb.from_dataframe()` fails

```
File /ocean/projects/phy210048p/malanche/lsdb-tests/cenv/lib/python3.11/site-packages/nested_pandas/nestedframe/core.py:58, in NestedFrame._is_known_hierarchical_column(self, colname)
     56 """Determine whether a string is a known hierarchical column name"""
     57 if "." in colname:
---> 58     left, right = colname.split(".")
     59     if left in self.nested_columns:
     60         return right in self.all_columns[left]

ValueError: too many values to unpack (expected 2)
```

We probably should count dots or unpack it safer, e.g. with `left, right, *others = colname.split(".")`

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a descriprion of what I expected instead.
- [x] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

dougbrn:
issue pertains to query wrapping (adding for search discoverability)

gitosaurus:
Note that this particular bug appears to have been fixed by @dougbrn in commit 1cde0c10f8985ad6b6da468edeba5b3263f89a18.  #177 changed the method, and didn't regress.

PR #150, opened 2024-10-09
Use custom resolver for query and eval with nested frames.

gitosaurus:
Verify preflighting of nested expressions using AST visitation.

Remove logic for splitting queries by string.  Now the evaluation is handled by a nested column resolver, and the mixed-mode expressions are preflighted by examining the parsed abstract syntax tree for the query expression.

Resolves https://github.com/lincc-frameworks/nested-pandas/issues/59.

## Change Description

## Code Quality

### Bug Fix Checklist


src/nested_pandas/nestedframe/core.py
```diff
@@ -13,7 +13,26 @@
 from nested_pandas.series import packer
 from nested_pandas.series.dtype import NestedDtype
 
-from .utils import _ensure_spacing
+from ..series.packer import pack_sorted_df_into_struct
+from .utils import NestingType, check_expr_nesting
+
+
+class NestedSeries(pd.Series):
```
hombit, line 20: Is it actually "FlatSeries"? Or maybe `_FlatSeries` because it is not part of the public interface?

hombit (commented):
Thank you!

I'd ask for more testing and not-implemented error-handling. For example, I'm not sure that assignment with `eval` is supported: `.eval("nested.c = nested.a + nested.b")` or `.eval("nested.b = x + nested.a").`

It looks like it may also close #144 and #146? If yes, could you please add tests for covering those issues?

src/nested_pandas/nestedframe/core.py
```diff
@@ -363,25 +459,62 @@ def query(self, expr) -> Self:  # type: ignore[name-defined] # noqa: F821
...
+
+            def __getattr__(self, item_name: str):
+                if item_name in self._flat_nest:
+                    result = NestedSeries(self._flat_nest[item_name])
+                    # Assigning these properties directly in order to avoid any complication
+                    # or interference with the inherited pd.Series constructor.
+                    result.nest_name = self._nest_name
+                    result.flat_nest = self._flat_nest
+                    return result
+                raise AttributeError(f"No attribute {item_name}")
```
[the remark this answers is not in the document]
gitosaurus: Ah, good point.  I defined it there to keep it close to the point of use, but there's no reason to incur any memory or compute overhead.  Will move it up.

src/nested_pandas/nestedframe/core.py
gitosaurus: Good question.  While technically it does cache the flattened nest, it does retain a reference to the original nest name and its nest siblings, so it knows that it is part of a nest.  I do think it should be public and not private because it does sometimes become the return value of `NestedFrame.eval`, depending on the expression.

gitosaurus:
> Thank you!
> 
> I'd ask for more testing and not-implemented error-handling. For example, I'm not sure that assignment with `eval` is supported: `.eval("nested.c = nested.a + nested.b")` or `.eval("nested.b = x + nested.a").`
> 
> It looks like it may also close #144 and #146? If yes, could you please add tests for covering those issues?

Yes, this PR does resolve #146, thanks for the tip!  #144 is not resolved by this PR, but perhaps this same approach could be extended to cover it in a separate PR.

Good catch on the assignment problem.  The existing ability for `eval()` to insert base columns still works, but you're right, it can't insert into the nest yet.  I'll add those tests and see if adding nested columns can be done within the scope of this PR.  If not, I'll at least make a clearer error.

src/nested_pandas/nestedframe/core.py
gitosaurus: Makes sense.  I'll make it private and give it a name that doesn't suggest generality.  Maybe `_SeriesFromNest`?

hombit: Sounds good, thank you!

src/nested_pandas/nestedframe/core.py
```diff
@@ -13,7 +13,52 @@
...
+
+    @property
+    def _constructor(self) -> Self:  # type: ignore[name-defined] # noqa: F821
+        return _SeriesFromNest
+
+    @property
+    def _constructor_expanddim(self) -> Self:  # type: ignore[name-defined] # noqa: F821
+        return NestedFrame
+
+    __pandas_priority__ = 3500
```
dougbrn, line 35: Would you mind adding a code comment about what setting the priority does here?

src/nested_pandas/nestedframe/core.py
```diff
@@ -289,38 +333,110 @@ def from_lists(cls, df, base_columns=None, list_columns=None, name="nested"):
...
+            operation inplace and mutate the existing DataFrame. Otherwise,
+            a new DataFrame is returned.
+        **kwargs
+            See the documentation for :func:`eval` for complete details
+            on the keyword arguments accepted by
+            :meth:`~pandas.DataFrame.query`.
+
+        Returns
+        -------
+        ndarray, scalar, pandas object, or None
```
dougbrn, line 365: should this have "nested-pandas object" as well?

src/nested_pandas/nestedframe/core.py
```diff
@@ -289,38 +333,110 @@ def from_lists(cls, df, base_columns=None, list_columns=None, name="nested"):
...
+        DataFrame.query : Evaluates a boolean expression to query the columns
+            of a frame.
+        DataFrame.assign : Can evaluate an expression or function to create new
+            values for a column.
+        eval : Evaluate a Python expression as a string using various
+            backends.
+
+        Notes
+        -----
+        For more details see the API documentation for :func:`~eval`.
```
dougbrn, line 379: The "See Also" and "Notes" sections here are referencing the Pandas docs, most of the links won't work as a result. I'd recommend either removing these, or replacing them with a note pointing to https://pandas.pydata.org/docs/reference/api/pandas.eval.html

src/nested_pandas/nestedframe/core.py
```diff
@@ -363,25 +485,38 @@ def query(self, expr) -> Self:  # type: ignore[name-defined] # noqa: F821
...
-        exprs_to_use = self._split_query(expr)
-
-        # For now (simplicity), limit query to only operating on one layer
-        if len(exprs_to_use.keys()) != 1:
+        if not isinstance(expr, str):
+            msg = f"expr must be a string to be evaluated, {type(expr)} given"
+            raise ValueError(msg)
+        kwargs["level"] = kwargs.pop("level", 0) + 1
+        kwargs["target"] = None
+        # At present, the query expression must be either entirely within the
```
dougbrn, line 493: Totally fine to keep this behavior, but if you believe this new implementation would support multi-target queries this could be a restriction we'd lift. Maybe opening a separate issue for this would be best

dougbrn (commented):
Thanks for this @gitosaurus , looks really good! I added a few comments, in addition to the feedback Kostya has provided.

src/nested_pandas/nestedframe/core.py
gitosaurus: I believe this PR makes a significant step toward multi-target queries, but it might be better to make the next step in a separate PR, for simplicity's sake.  Will investigate.

gitosaurus:
> > Thank you!
> > I'd ask for more testing and not-implemented error-handling. For example, I'm not sure that assignment with `eval` is supported: `.eval("nested.c = nested.a + nested.b")` or `.eval("nested.b = x + nested.a").`
> > It looks like it may also close #144 and #146? If yes, could you please add tests for covering those issues?
> 
> Yes, this PR does resolve #146, thanks for the tip! #144 is not resolved by this PR, but perhaps this same approach could be extended to cover it in a separate PR.
> 
> Good catch on the assignment problem. The existing ability for `eval()` to insert base columns still works, but you're right, it can't insert into the nest yet. I'll add those tests and see if adding nested columns can be done within the scope of this PR. If not, I'll at least make a clearer error.

The changes I pushed yesterday do permit assignment of nested columns, and even the creation of new nests, from within `eval()`!  Added tests of this as well.

src/nested_pandas/nestedframe/core.py
gitosaurus: Given how long this PR has been in flight, and the amount of complexity in it, I will leave the multi-target queries for a future PR that builds on top of this one.

tests/nested_pandas/nestedframe/test_nestedframe.py
```diff
@@ -689,3 +734,164 @@ def cols_allclose(col1, col2):
...
+    Test the ability to use expected methods and functions within eval(),
+    on nested columns.
+    """
+    # Verifies https://github.com/lincc-frameworks/nested-pandas/issues/146
+    nf = NestedFrame.from_flat(NestedFrame({"a": [1, 2], "b": [3, None]}, index=[1, 1]), base_columns=[])
+    assert nf["nested.b"].shape == (2,)
+    assert nf.query("nested.b.isna()")["nested.b"].shape == (1,)
+
+    assert nf["nested.a"].max() == nf.eval("nested.a.max()") == 2
+    assert nf["nested.a"].min() == nf.eval("nested.a.min()") == 1
```
hombit, line 793: I LOVE IT

tests/nested_pandas/nestedframe/test_nestedframe.py
```diff
@@ -689,3 +734,164 @@ def cols_allclose(col1, col2):
...
+        index=pd.Index([0, 0, 0, 1, 1, 1, 2, 2, 2], name="idx"),
+    )
+    nf = nf.add_nested(to_pack, "packed")
+    # Assigning to new base columns from old base columns
+    nf_b = nf.eval("c = a + 1")
+    assert len(nf_b.columns) == len(nf.columns) + 1
+    assert (nf_b["c"] == nf["a"] + 1).all()
+
+    # Assigning to new nested columns from old nested columns
+    nf_nc = nf.eval("packed.e = packed.c + 1")
```
hombit, line 848: OMG, IT IS SO GOOD

hombit (approved):
Great work, thank you so much!

dougbrn (approved):
This is great! I think the eval interface here is a huge addition to nested-pandas, to the level that I would love to see it in our documentation. I wouldn't ask it for this PR, but perhaps in a future PR we could add a section to the [data manipulation](https://nested-pandas.readthedocs.io/en/latest/tutorials/data_manipulation.html) tutorial showcasing some of the uses of eval?

gitosaurus:
> This is great! I think the eval interface here is a huge addition to nested-pandas, to the level that I would love to see it in our documentation. I wouldn't ask it for this PR, but perhaps in a future PR we could add a section to the [data manipulation](https://nested-pandas.readthedocs.io/en/latest/tutorials/data_manipulation.html) tutorial showcasing some of the uses of eval?

Great idea!  I'll add a task for that.

src/nested_pandas/nestedframe/core.py
```diff
@@ -9,11 +10,123 @@
...
+                raise KeyError(f"Unknown nest {top_nest}")
+            super().__setitem__(top_nest, _NestedFieldResolver(top_nest, self._outer))
+        return super().__getitem__(top_nest)
+
+    def __setitem__(self, key, value):
+        # Called to update the resolver with intermediate values.
+        # The important point is to intercept the call so that the evaluator
+        # does not create any new resolvers on the fly.  Storing the value
+        # is not important, since that will have been done already in
+        # the NestedFrame.
```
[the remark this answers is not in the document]
gitosaurus: Indeed the `eval` machinery does give a resolver the opportunity to update its value for a given key; the motive seems to be to permit resolvers to hold the intermediate values for multi-line evaluation.  However, in the case of `NestedFrame`, the resolver is, intentionally, a lazy and on-demand resolver, which means that by the time `eval` provides the value to the assigner-tag, the `NestedFrame` already knows about it.

Moreover, it's important that the nested resolver understand that an expression like `nested.a` is addressing the entity `nested`, something that the default eval machinery doesn't assume.  For both of these reasons, the `__setitem__` functionality of this particular resolver is not merely unnecessary but could create confusion.

src/nested_pandas/nestedframe/core.py
```diff
@@ -289,38 +405,48 @@ def from_lists(cls, df, base_columns=None, list_columns=None, name="nested"):
...
+            The result of the evaluation or None if ``inplace=True``.
+
+        See Also
+        --------
+        https://pandas.pydata.org/docs/reference/api/pandas.eval.html
+        """
+        kwargs["resolvers"] = tuple(kwargs.get("resolvers", ())) + (_NestResolver(self),)
+        kwargs["inplace"] = inplace
+        kwargs["parser"] = "nested-pandas"
+        return super().eval(expr, **kwargs)
```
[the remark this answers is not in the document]
gitosaurus: It can return `_SeriesFromNest` on occasion, true.  This is a descendant of `pd.Series` already, so expressions like `instanceof(obj, pd.Series)` will be true for it.  I can't think of an easy way to change it to `pd.Series` (and nothing more) without impacting the ability for `query` to do the right thing; it uses this type tag to decide what level of querying to perform.

PR #151, opened 2024-10-09
Update PPT to latest version

jeremykubica:
Mechanical application of PPT update


src/nested_pandas/example_benchmarks.py
dougbrn, line 1: We should remove this file from the PR

docs/notebooks/README.md
dougbrn, line 1: PPT generated a notebooks directory (which we renamed to tutorials in nested-pandas), can we remove this directory from the PR

docs/pre_executed/README.md
dougbrn, line 1: Same for above, removing this would be good

dougbrn (changes requested):
Looks good, just requesting a few things be not added to the repository

src/nested_pandas/example_benchmarks.py
jeremykubica: Good call. I will remove this (and the one from nested-dask as well)

docs/notebooks/README.md
jeremykubica: Good call. I will remove this (and the one from nested-dask as well)

docs/pre_executed/README.md
jeremykubica: Good call. I will remove this (and the one from nested-dask as well)

PR #162, opened 2024-10-29
Invalidate nest-resolver caches during assignment.

### LINKED ISSUE #159: NestedFrame.eval does not recognize new columns in existing nests in multiline strings
**Bug report**

**Before submitting**
Please check the following:

- [X] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.

When `NestedFrame.eval` is used with multiline strings to create new columns in existing nests, it fails to recognize these new columns later within the string:

```python
nf = generate_data(10, 100)                                                                                                                               
nf.eval("""                                                                                                                                               
    c = a + b                                                                                                                                             
    nested.d = nested.flux * 2                                                                                                                            
    n2.gumption = nested.d / (a + b)                                                                                                                      
    """, inplace=False)  # value of inplace does not matter                                                                                               
```

`nested.d` is not recognized.  It will be recognized on the *next* call to `NestedFrame.eval`, but not within this string.

This is because the `_NestedFieldResolver` used by the evaluator caches the flattened nest upon referencing it, so that it doesn't notice changes to it.  Note that the changes really have been made: `nested.d` is created, just not visible to the later lines in the multiline string.

The solution is to invalidate the appropriate `_NestedFieldResolver` instance's cache when an `eval()` assignment is made.  The best place to do this is probably in `_NestResolver.__setitem__`.

gitosaurus:
Closes #159 .

## Change Description

## Code Quality

### Bug Fix Checklist

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -94,13 +94,19 @@ class _NestResolver(dict):
             super().__setitem__(top_nest, _NestedFieldResolver(top_nest, self._outer))
         return super().__getitem__(top_nest)
 
-    def __setitem__(self, key, value):
+    def __setitem__(self, item, _):
         # Called to update the resolver with intermediate values.
         # The important point is to intercept the call so that the evaluator
-        # does not create any new resolvers on the fly.  Storing the value
-        # is not important, since that will have been done already in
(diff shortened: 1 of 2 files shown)


wilsonbb (approved):
LGTM!

PR #165, opened 2024-10-31
Query expression preflighting now prohibits crossing nests.

### LINKED ISSUE #160: Cross-nest queries in NestedFrame.query should be prohibited or split
**Bug report**

**Before submitting**
Please check the following:

- [X] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.

Processing queries across base and nested columns is not possible in `NestedFrame.query` today, so expressions of this kind are detected and prohibited.  However, it does not presently prohibit queries that go across two different nests:

```python
nf.query("n1.c < 0.5 and n2.d > 1e-7")
```

The above will "work", but only selects the nested rows in `n1`, simply because `n1` is earlier in the string.

The solution is to treat such cross-nest queries exactly the same as queries that cross base and nested columns: error out today, and as split queries, in the future.

gitosaurus:
Doing so never worked, but now it is correctly detected.

Resolves #160 .

## Change Description

## Solution Description

The expression-tree scanning that was done before is now expanded so that it not only notices the use of nests, but tracks the nest columns' identifiers.

## Code Quality

### Bug Fix Checklist

### DIFF
--- a/src/nested_pandas/nestedframe/utils.py
+++ b/src/nested_pandas/nestedframe/utils.py
@@ -2,14 +2,6 @@
 from __future__ import annotations
 
 import ast
-from enum import Enum
-
-
-class NestingType(Enum):
-    """Types of sub-expressions possible in a NestedFrame string expression."""
-
-    BASE = "base"
-    NESTED = "nested"
 
 
 def _actionable_splits(parents: list[ast.expr], node: ast.expr | None) -> dict[str, list]:
@@ -48,11 +40,10 @@ def _actionable_splits(parents: list[ast.expr], node: ast.expr | None) -> dict[s
         # this node is homogeneous over all of its children, and can
         # be evaluated in a single step.
         result = {k: [node] for k in result}
-    else:
-        # At this point, we need to split the expression.  The idea here is that
-        # we want a succession of efficient queries, each of which will produce
-        # a subset of either the base or the nested columns.
-        pass
+    # If the result is either empty or has more than one key, leave the result
+    # alone.  Each key represents a different nest (with a blank string for the base),
+    # and the value is the highest point in the expression tree where the expression
+    # was still within a single nest.
     return result
 
 
--- a/src/nested_pandas/nestedframe/utils.py
+++ b/src/nested_pandas/nestedframe/utils.py
@@ -4,17 +4,20 @@ from __future__ import annotations
 import ast
 
 
-def _actionable_splits(parents: list[ast.expr], node: ast.expr | None) -> dict[str, list]:
+def _subexprs_by_nest(parents: list[ast.expr], node: ast.expr | None) -> dict[str, list]:
     """
-    Given an expression which contains references to both base and nested columns,
-    return a dictionary of the sub-expressions that should be evaluated independently.
-    The key of the dictionary is the name of the nested column, and will be a blank
-    string in the case of base columns.  The value is a list of the parent nodes
-    that lead to sub-expressions that can be evaluated successfully.
+    Given an expression which contains references to both base and nested
+    columns, return a dictionary of the sub-expressions that should be
+    evaluated independently, keyed by nesting context.
 
-    While this is not in use today for automatically splitting expressions, it can
-    be used to detect whether an expression is suitably structured for evaluation:
-    the returned dictionary should have a single key.
+    The key of the dictionary is the name of the nested column, and will
+    be a blank string in the case of base columns.  The value is a list
+    of the parent nodes that lead to sub-expressions that can be evaluated
+    successfully.
+
+    While this is not in use today for automatically splitting expressions,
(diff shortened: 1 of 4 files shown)


src/nested_pandas/nestedframe/utils.py
```diff
@@ -12,29 +12,56 @@ class NestingType(Enum):
     NESTED = "nested"
```
wilsonbb, line 12: Is the NestingType enum still in use? At a glance it looks like we've removed all references to it?

src/nested_pandas/nestedframe/utils.py
```diff
@@ -12,29 +12,56 @@ class NestingType(Enum):
...
+    else:
+        # At this point, we need to split the expression.  The idea here is that
+        # we want a succession of efficient queries, each of which will produce
+        # a subset of either the base or the nested columns.
+        pass
     return result
 
 
-def check_expr_nesting(expr: str) -> set[NestingType]:
+def check_expr_nesting(expr: str) -> set[str]:
```
wilsonbb, line 59: Nit: It feels like it would be nice to rename this function since it is no longer returning the set of checked types

Maybe something like `extract_expr_nesting` or `extract_nested_expr`

But I'm not particularly good at naming things

src/nested_pandas/nestedframe/utils.py
```diff
@@ -12,29 +12,56 @@ class NestingType(Enum):
...
+        # Let the record of each parent node drift up the tree,
+        # and merge the subtrees into a single node, since by definition,
+        # this node is homogeneous over all of its children, and can
+        # be evaluated in a single step.
+        result = {k: [node] for k in result}
+    else:
+        # At this point, we need to split the expression.  The idea here is that
+        # we want a succession of efficient queries, each of which will produce
+        # a subset of either the base or the nested columns.
+        pass
```
wilsonbb, line 55: This is currently a no-op with confusing documentation. Do we still need this code block or was this a remnant from a different way of constructing the recursion?

wilsonbb (commented):
Overall it looks pretty good, but I am a bit confused by a part of _actionable_splits. Can check again in the morning.

src/nested_pandas/nestedframe/utils.py
```diff
@@ -12,29 +12,56 @@ class NestingType(Enum):
...
     """
     Given a string expression, parse it and visit the resulting AST, surfacing
     the nesting types.  The purpose is to identify expressions that attempt
-    to mix base and nested columns, which will need to be handled specially.
+    to mix base and nested columns, or columns from two different nests.
     """
     expr_tree = ast.parse(expr, mode="eval").body
-    return _expr_nesting_type(expr_tree)
+    separable = _actionable_splits([], expr_tree)
+    return set(list(separable.keys()))
```
hombit, line 67: Set can be built from iterable
```suggestion
    return set(separable.keys())
```

src/nested_pandas/nestedframe/utils.py
```diff
@@ -12,29 +12,56 @@ class NestingType(Enum):
...
         + getattr(node, "values", [])
         + getattr(node, "comparators", [])
     )
-    result: set[NestingType] = set()
+    result: dict[str, list] = {}
     for s in sources:
-        result.update(_expr_nesting_type(s))
+        child = _actionable_splits(parents, s)
+        for k, v in child.items():
+            result[k] = result.get(k, []) + v
```
hombit, line 42: It may be a matter of taste, feel free to reject this suggestion:
```suggestion
            result.setdefault(k, []).append(v)
```

hombit (approved):
Could you please add a higher level test checking that an example from #160 fails?

src/nested_pandas/nestedframe/utils.py
gitosaurus: You're right, the enum is not longer used and should be removed.  Will do so.

src/nested_pandas/nestedframe/utils.py
gitosaurus: Can do.  What it's doing now is identifying which nests are referenced in the given expression, with base represented by the empty string.  🤔  Maybe `identify_nests_in_expr` or `extract_nest_names`.  I kind of like the latter.

src/nested_pandas/nestedframe/utils.py
gitosaurus: Good eye!  It is indeed a no-op, but the no-op is very intentional, hence the comment.  You're right, this was scaffolding for a more ambitious future implementation, which I scaled back.  I'll restructure it to lose the `else / pass`, and make that comment more pertinent to this more concise algorithm.

gitosaurus:
> Could you please add a higher level test checking that an example from #160 fails?

Done!

PR #168, opened 2024-11-01
Handle non-unique index

hombit:


src/nested_pandas/nestedframe/core.py
```diff
@@ -518,18 +519,37 @@ def query(self, expr: str, *, inplace: bool = False, **kwargs) -> NestedFrame |
...
+                raise ValueError("Query condition must evaluate to a boolean Series")
+
             nest_name, flat_nest = result.nest_name, result.flat_nest
-            new_flat_nest = flat_nest.loc[result]
-            result = self.copy()
-            result[nest_name] = pack_sorted_df_into_struct(new_flat_nest)
+
+            # Reset index to "ordinal" like [0, 0, 0, 1, 1, 2, 2, 2]
+            flat_nest = flat_nest.set_index(self[nest_name].array.list_index)
+            query_result = result.set_axis(self[nest_name].array.list_index)
```
gitosaurus, line 529: This property `self[nest_name].array.list_index` does involve a (small) amount of computation; should it be pulled into a variable in order to be done once?  Or is the goal here to ensure that there are two distinct objects?

src/nested_pandas/nestedframe/core.py
```diff
@@ -518,18 +519,37 @@ def query(self, expr: str, *, inplace: bool = False, **kwargs) -> NestedFrame |
         # to the nest and repack.  Otherwise, apply it to this instance as usual,
         # since it operated on the base attributes.
         if isinstance(result, _SeriesFromNest):
+            if not is_bool_dtype(result.dtype):
+                raise ValueError("Query condition must evaluate to a boolean Series")
```
gitosaurus, line 523: Adding a simple test to trigger this has yielded some odd and unexpected errors:

```python
base.query("a + 2 / b")
```
does not yield this `ValueError`, but a `KeyError` complaining that some array (it shows the values) does not exist in `[index]`.  Still looking into this, but it appears some index plumbing is amiss.

src/nested_pandas/nestedframe/core.py
```diff
@@ -219,10 +218,8 @@ def __setitem__(self, key, value):
             "." in key and key.split(".")[0] in self.nested_columns
         ):
             nested, col = key.split(".")
-            new_flat = self[nested].nest.to_flat()
-            new_flat[col] = value
-            packed = packer.pack(new_flat)
-            return super().__setitem__(nested, packed)
+            new_nested_series = self[nested].nest.with_flat_field(col, value)
```
gitosaurus, line 221: If I understand, this change avoids rewriting fields that haven't changed, true?

src/nested_pandas/series/ext_array.py
```diff
@@ -648,6 +648,12 @@ def num_chunks(self) -> int:
         """Number of chunks in underlying pyarrow.ChunkedArray"""
         return self._chunked_array.num_chunks
 
+    @property
+    def list_index(self) -> np.ndarray:
+        """Keys mapping values to lists"""
```
gitosaurus, line 653: This docstring is short enough to be a little puzzling; there's a lot going on in a few statements here.

src/nested_pandas/nestedframe/core.py
hombit: We can make it a method and pull to a variable. Could you please do it if you have some time?

src/nested_pandas/series/ext_array.py
hombit: Name is also not great, I'd appreciate a help with renaming

hombit: The idea was this index mapping "flat value" to "list value"

src/nested_pandas/nestedframe/core.py
hombit: Yes, and benchmarks show that it really helps

src/nested_pandas/nestedframe/core.py
hombit: We don’t want to support this index functionality, at least not right now. Thank you for investigating to provide a better error message!

gitosaurus: Verified just now that this same problem afflicts vanilla Pandas DataFrame objects, so it's not related to our changes and is probably out of scope here.  It does make it harder to trigger the new `ValueError`; maybe that's what we should scale back.

tests/nested_pandas/nestedframe/test_nestedframe.py
```diff
@@ -212,6 +217,35 @@ def test_add_nested_with_flat_df_and_mismatched_index():
...
+        left_res_on = base.add_nested(nested, "nested", how="left", on=["new_index"])
+    # Instead we should pass a single column name, "new_index" which exists in both frames.
+    left_res_on = base.add_nested(nested, "nested", how="left", on="new_index")
+    assert "nested" in left_res_on.columns
+    # Check that the index of the base layer is still being used
+    assert (left_res_on.index == base.index).all()
+    # Assert that the new_index column we joined on was dropped from the nested layer
+    # but is present in the base layer
+    assert "new_index" in left_res_on.columns
+    assert "new_index" not in left_res_on["nested"].nest.to_flat().columns
```
gitosaurus, line 233: Same as checking within `.nest.fields`, yes?  Does going through `.to_flat()` test this better?

gitosaurus (approved):

wilsonbb (approved):
Looks good to me! Minor suggestions including fixing one of my own typos haha

tests/nested_pandas/nestedframe/test_nestedframe.py
hombit: CC @wilsonbb

hombit: I think both are fine, let's keep it as is for now?

wilsonbb: Yeah that's fine by me

PR #169, opened 2024-11-05
Tests for using the 'on' keyword with add_nested and packing

wilsonbb:
Adds unit tests using the 'on' keyword in `NestedFrame.add_nested` and packer methods. Based off of the branch `non-uniq-idx`

We also modify the unit test for `NestedFrame.query` to be performed on a `NestedFrame` with a non-unique index and a simple unit test for the new `get_list_index` method.

It also incorporates minor fixes from @gitosaurus in merged in https://github.com/lincc-frameworks/nested-pandas/pull/171

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -271,9 +271,8 @@ class NestedFrame(pd.DataFrame):
               index, and sort it lexicographically.
             - inner: form intersection of calling frame's index with other
               frame's index, preserving the order of the calling index.
-        on : str, list of str, default: None
-            Columns in `obj` frame to use as an index to join on rather than
-            `obj`'s index.
+        on : str, default: None
+            A column in the list 
         dtype : dtype or None
             NestedDtype to use for the nested column; pd.ArrowDtype or
             pa.DataType can also be used to specify the nested dtype. If None,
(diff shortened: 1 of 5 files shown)


src/nested_pandas/series/ext_array.py
```diff
@@ -648,9 +648,11 @@ def num_chunks(self) -> int:
         """Number of chunks in underlying pyarrow.ChunkedArray"""
         return self._chunked_array.num_chunks
 
-    @property
-    def list_index(self) -> np.ndarray:
+    def get_list_index(self) -> np.ndarray:
         """Keys mapping values to lists"""
+        if len(self) == 0:
+            # Since we have no list offests, return an empty array
+            return np.empty(0)
```
hombit, line 655:
```suggestion
            return np.empty(0, dtype=int)
```
or 
```suggestion
            return np.array([], dtype=int)
```

src/nested_pandas/nestedframe/core.py
```diff
@@ -271,8 +270,8 @@ def add_nested(
               index, and sort it lexicographically.
             - inner: form intersection of calling frame's index with other
               frame's index, preserving the order of the calling index.
-        on : str, list of str, default: None
-            Columns to join on.
+        on : str, default: None
+            A column in the list
```
hombit, line 274:
```suggestion
            A column name to nest and join on
```

hombit (approved):
Looks great

PR #175, opened 2024-11-14
Use the Pandas expression tree for query preflighting.

### LINKED ISSUE #174: Fix NestedFrame.query/eval to handle column names including a space
We recently introduced an AST parser for evaluation of expressions against Nested-Pandas columns in `eval` and `query`.

However this seems to fail against column names which include a space (which is allowed in base Pandas).

This was discovered when updating the `nested-dask` version for `lsdb` since one of their unit tests exercises a query against such a column. Here is the example test failure with a full stack trace when our current query/eval parser is used by LSDB: https://github.com/astronomy-commons/lsdb/actions/runs/11713280862/job/32625751711?pr=493

A pretty trivial reproduction:

```
from nested_pandas import NestedFrame
import pandas as pd

nf = NestedFrame(data={"dog": [1, 2, 3], "good dog": [2, 4, 6]}, index=[0, 1, 2])

nested = pd.DataFrame(
    data={"a": [0, 2, 4, 1, 4, 3, 1, 4, 1], "b": [5, 4, 7, 5, 3, 1, 9, 3, 4]},
    index=[0, 0, 0, 1, 1, 1, 2, 2, 2],
)

nf = nf.add_nested(nested, "nested")
nf.query("`good dog` > 3")
```

gitosaurus:
Requires `extract_nest_names` to be a method on `NestedFrame` so that the evaluation context is available at parsing time, since the Pandas `Expr` parsing does some eager evaluation.

Resolves #174 .

## Change Description

## Solution Description

The preflighting for `NestedFrame.query()` was examining the Python AST parse tree to locate the names of nests whose fields were being used in expressions.  This was fine so long as the column names were identifier-like, but failed for columns with spaces.  Pandas provides a way of dealing with such names, by enclosing them in backticks.  This causes Pandas to transform them into a temporary identifier that is put into a resolver and used for the duration; this process is called cleaning.  `.query()` needed to use this evaluator.

However, the Pandas evaluator does some eager evaluation, so that its parsing cannot occur outside of its `DataFrame` (and `NestedFrame`) context, so the proper nested field resolvers need to be preloaded for existing columns, as well as created on the fly for dynamic columns.

## Code Quality

### Bug Fix Checklist


hombit (approved):
Great! Probably it is out of scope of this PR, feel free to separate it into a new issue. We also need to support arbitrary names for sub-columns, for example this code doesn't work:

```python
nf = NestedFrame(data={"dog": [1, 2, 3], "good dog": [2, 4, 6]}, index=[0, 1, 2])
nested = pd.DataFrame(
    data={"n/a": [0, 2, 4, 1, 4, 3, 1, 4, 1], "n/b": [5, 4, 7, 5, 3, 1, 9, 3, 4]},
    index=[0, 0, 0, 1, 1, 1, 2, 2, 2],
)
nf = nf.add_nested(nested, "bad dog")
nf.query("`bad dog`.`n/a` > 2")
```

It is nice that nested column name could be arbitrary now, but if we need to choose between supporting arbitrary nested column names and sub-column names, I definitely would prefer the second option. The reason is that nested column name is under our control, while sub-columns are coming from user's data.

wilsonbb (commented):
LGTM!

PR #186, opened 2025-01-07
Expand quickstart.ipynb

jeremykubica:
Add some additional explanation and examples to the quick start notebook.

### DIFF
--- a/docs/gettingstarted/quickstart.ipynb
+++ b/docs/gettingstarted/quickstart.ipynb
@@ -20,7 +20,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -296,7 +296,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 15,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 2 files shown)


docs/gettingstarted/quickstart.ipynb
```diff
@@ -4,13 +4,17 @@
    "cell_type": "markdown",
```
dougbrn, line 4: <p>I really like the lead-in you've added here. One thing I'm noticing however is that it's a little jarring that the nested dataframe example here isn't the same toy dataset from the lead-in. It would be great if the toy dataset could be changed here. Potentially by way of <code>from_flat</code>:</p><p> </p><p><code>nf = NestedFrame.from_flat(my_data_frame, base_columns=["id", "ra","dec"])  # or set the id as index and don't use as base_column</code></p><p> </p><p>All downstream column references would need to be changed as well (or we could switch to the dummy when we start showing off query/reduce?), but this may be nice as it also shows users how to turn their data into nestedframes right away.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/nested-pandas/pull/186/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

jeremykubica: That's fair. I rewrote the notebook to use the astronomy data as a running example. Let me know what you think.

dougbrn (approved):
Looks great, thank you!

PR #197, opened 2025-02-21
Fix NestedExtenstionArray.set_flat_field for chunked values

hombit:
Another small bug fix for chunked arrays


gitosaurus (approved):
LGTM

PR #198, opened 2025-02-21
List lengths property for extension array and accessor

hombit:
I believe it is very useful to have, the alternative is calling something like
```python
ndf[nested].nest.to_lists()["some_column_name"].list.len()
```

### DIFF
--- a/src/nested_pandas/series/accessor.py
+++ b/src/nested_pandas/series/accessor.py
@@ -115,6 +115,11 @@ class NestSeriesAccessor(Mapping):
         return pd.DataFrame(flat_series)
 
     @property
+    def list_lengths(self) -> np.ndarray:
+        """Lengths of the list arrays"""
+        return self._series.array.list_lengths
+
+    @property
     def flat_length(self) -> int:
         """Length of the flat arrays"""
         return self._series.array.flat_length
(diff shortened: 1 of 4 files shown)


src/nested_pandas/series/accessor.py
```diff
@@ -114,6 +114,11 @@ def to_flat(self, fields: list[str] | None = None) -> pd.DataFrame:
 
         return pd.DataFrame(flat_series)
 
+    @property
+    def list_lengths(self) -> np.ndarray:
```
gitosaurus, line 118: This says it returns a numpy array, but it looks like the property it references is a `list[int]`, unless I'm missing something

tests/nested_pandas/series/test_accessor.py
```diff
@@ -286,6 +286,19 @@ def test_fields():
...
+    """Test that the .nest.list_lengths attribute works."""
+    series = pack_seq(
+        [
+            pd.DataFrame({"a": [1, 2, 3], "b": [1.0, 5.0, 6.0], "c": ["a", "b", "c"]}),
+            None,
+            pd.DataFrame({"a": [1, 2], "b": [None, 0.0], "c": ["a", "b"]}),
+        ]
+    )
+    assert series.shape == (3,)
+    assert series.nest.list_lengths == [3, 0, 2]
```
gitosaurus, line 299: This seems to confirm `list[int]`

src/nested_pandas/series/accessor.py
gitosaurus: To be clear, I think that `list[int]` is the right return type, and this type marker should change to match

hombit: Nice catch!

Issue #202, opened 2025-02-24
Implemented ndf.sort_values for nested columns


hombit:
**Feature request**

It would be really nice if we would have a simple and performant way to sort values inside nested frames. I think that we can reimplement `sort_values` allowing subcolumns with dot-notation. To not confuse user I'd allow single nested column per call.

Allow:
- `ndf.sort_values("a")`
- `ndf.sort_values(["a", "b"])`
- `ndf.sort_values("lc.time")`
- `ndf.sort_values(["lc.band", "lc.time"])`

Do not allow:
- `ndf.sort_values(["a", "lc.time"])`
- `ndf.sort_values(["detections.time", "non_detections.band"])`

The implementation would be something like this:
```python
flat = ndf.lc.nest.to_flat()
flat = flat.sort_values([flat.index.name] + sort_subcolumns)
ndf.lc[:] = flat
```

**Before submitting**
Please check the following:

- [x] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [x] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [x] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

PR #205, opened 2025-02-26
Wrapper for DataFrame.sort_values

dougbrn:
Addresses #202

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -994,7 +994,7 @@ class NestedFrame(pd.DataFrame):
                 ignore_index=ignore_index,
                 key=key,
             )
-        elif target in self.nested_columns:
+        else:  # target is a nested column
             target_flat = self[target].nest.to_flat()
             target_flat = target_flat.set_index(self[target].array.get_list_index())
 
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -1003,6 +1003,12 @@ class NestedFrame(pd.DataFrame):
             # Index must always be the first sort key for nested columns
             nested_by = [target_flat.index.name] + [col.split(".")[-1] for col in by]
 
+            # Augment the ascending kwarg to include the index
+            if isinstance(ascending, bool):
+                ascending = [True] + [ascending] * len(by)
+            elif isinstance(ascending, list):
+                ascending = [True] + ascending
+
             if inplace:
                 target_flat.sort_values(
                     by=nested_by,
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -1009,28 +1009,16 @@ class NestedFrame(pd.DataFrame):
             elif isinstance(ascending, list):
                 ascending = [True] + ascending
 
-            if inplace:
-                target_flat.sort_values(
-                    by=nested_by,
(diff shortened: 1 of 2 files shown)


src/nested_pandas/nestedframe/core.py
```diff
@@ -916,6 +916,127 @@ def dropna(
...
+                by=by,
+                axis=axis,
+                ascending=ascending,
+                inplace=inplace,
+                kind=kind,
+                na_position=na_position,
+                ignore_index=ignore_index,
+                key=key,
+            )
+        elif target in self.nested_columns:
```
hombit, line 997: Do we need this `elif`? there is no `else` which would fail for unknown nested column after that. Could you please a test-case for that?

src/nested_pandas/nestedframe/core.py
```diff
@@ -916,6 +916,127 @@ def dropna(
...
+            nested_by = [target_flat.index.name] + [col.split(".")[-1] for col in by]
+
+            # Augment the ascending kwarg to include the index
+            if isinstance(ascending, bool):
+                ascending = [True] + [ascending] * len(by)
+            elif isinstance(ascending, list):
+                ascending = [True] + ascending
+
+            if inplace:
+                target_flat.sort_values(
```
hombit, line 1013: I'm not sure if `inplace` would actually change the nested series, could you please test it? Actually I think we should just sort it "out-of-place" and then the `if inplace` bellow would do what is needed

src/nested_pandas/nestedframe/core.py
```diff
@@ -916,6 +916,127 @@ def dropna(
...
+                    by=nested_by,
+                    axis=axis,
+                    ascending=ascending,
+                    kind=kind,
+                    na_position=na_position,
+                    ignore_index=False,
+                    key=key,
+                    inplace=False,
+                )
+            new_df = self._set_filtered_flat_df(nest_name=target, flat_df=target_flat)
```
hombit, line 1034: I don't think that we need this filtered helper function here, because we haven't change number of flat rows. I'd do something like following instead:
```python
            new_df = df.copy()
            new_df[target].array[:] = target_flat
```
We probably can optimize it for inplace case.

We do need some `.nest` method for full replacement of the underlying data with a provided flat-frame...

hombit (commented):
Thanks, please see few suggestions

src/nested_pandas/nestedframe/core.py
dougbrn: Originally had an else, but realized there was no way to actually reach that code practically. Switched this elif to just be an else as we effectively validate it's a real nested column to add it as a target layer (if we specify an invalid nested column it will target the base layer and error out using the pandas missing column error)

src/nested_pandas/nestedframe/core.py
dougbrn: Yeah good catch! I do verify in unit tests and in some additional notebook examples that the nested series is affected by the sorting inplace

src/nested_pandas/nestedframe/core.py
hombit: oof, sorry. Let's keep your code for now!

hombit (approved):

src/nested_pandas/nestedframe/core.py
dougbrn: No worries, will add a comment that we can potentially optimize this

Issue #208, opened 2025-02-27
html_repr fails when pd.get_option("display.max_rows")=None


dougbrn:
**Bug report**
Pointed out by @nevencaplar 

Reproduced by this code:
```
import nested_pandas as npd
import pandas as pd
pd.set_option("display.max_rows", None)
base = npd.NestedFrame(data={"a": [3, 2, 1], "b": [2, 3, 6]}, index=[0, 1, 2])
nested = pd.DataFrame(
    data={"c": [0, 2, 5, 1, 4, 3, 1, 4, 1], "d": [5, 4, 7, 5, 3, 1, 9, 3, 4]},
    index=[0, 0, 0, 1, 1, 1, 2, 2, 2],
)
ndf = base.add_nested(nested, "nested")
ndf
```

**Before submitting**
Please check the following:

- [ ] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [ ] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.
- [ ] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

Issue #207, opened 2025-02-27
html_repr doesn't handle dropped dataframes


dougbrn:
**Bug report**
Through filtering steps, we can get empty nested dataframes. These are replaced in the materialized view with just a NoneType, so the _repr_html code will fail when trying to call a to_html method from the nonetype. This is the reproducing code:
```
import nested_pandas as npd
import pandas as pd
base = npd.NestedFrame(data={"a": [3, 2, 1], "b": [2, 3, 6]}, index=[0, 1, 2])
nested = pd.DataFrame(
    data={"c": [0, 2, 5, 1, 4, 3, 1, 4, 1], "d": [5, 4, 7, 5, 3, 1, 9, 3, 4]},
    index=[0, 0, 0, 1, 1, 1, 2, 2, 2],
)
ndf = base.add_nested(nested, "nested")

ndf.query("nested.c>4")
```
And this is the error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File [~/.virtualenvs/lsdb/lib/python3.12/site-packages/IPython/core/formatters.py:406](http://localhost:8888/~/.virtualenvs/lsdb/lib/python3.12/site-packages/IPython/core/formatters.py#line=405), in BaseFormatter.__call__(self, obj)
    404     method = get_real_method(obj, self.print_method)
    405     if method is not None:
--> 406         return method()
    407     return None
    408 else:

File [~/lincc/nested-pandas/src/nested_pandas/nestedframe/core.py:319](http://localhost:8888/~/lincc/nested-pandas/src/nested_pandas/nestedframe/core.py#line=318), in NestedFrame._repr_html_(self)
    316     html_repr = repr.to_html(max_rows=pd.get_option("display.min_rows"))
    317 else:
    318     # when under the max_rows threshold, display all rows (behavior of 0 here)
--> 319     html_repr = repr.to_html(max_rows=0)
    321 # Manually append dimensionality to a styler output
    322 html_repr += f"{repr.data.shape[0]} rows x {repr.data.shape[1]} columns"

[...]

PR #209, opened 2025-02-27
Double Header: fix max_rows & empty chunk bugs

dougbrn:
Addresses #207 and #208.

No unit tests added as this is repr code, but verified that the reproducible examples in the issues now work.

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -308,8 +308,7 @@ class NestedFrame(pd.DataFrame):
             # If the chunk is None, just return None
             if chunk is None:
                 return None
-            else:
-                return chunk.to_html(max_rows=1, max_cols=5, show_dimensions=True, index=False, header=False)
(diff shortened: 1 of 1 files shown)


src/nested_pandas/nestedframe/core.py
```diff
@@ -301,7 +305,11 @@ def repack_first_cell(chunk):
 
         # remaining cells show only a preview row
         def repack_row(chunk):
-            return chunk.to_html(max_rows=1, max_cols=5, show_dimensions=True, index=False, header=False)
+            # If the chunk is None, just return None
+            if chunk is None:
+                return None
+            else:
+                return chunk.to_html(max_rows=1, max_cols=5, show_dimensions=True, index=False, header=False)
```
gitosaurus, line 312: Could drop `else` and outdent

gitosaurus (approved):

src/nested_pandas/nestedframe/core.py
dougbrn: Good call!

PR #213, opened 2025-03-05
Add methods to fill nested lists with repeated values

### LINKED ISSUE #200: Add tooling to embed a base column into a nested column
**Feature request**

It would be nice if we have a way to convert a base column to a field of a nested column, so the values are repeated for each row. Possible interface:
```python
ndf = generate_data(100, 10)
ndf.add_repeated_nested(ndf["a"], name="nested.a")
# Now ndf["nested"][0] has "a" column and all the values are equal to ndf["a"][0]
ndf.add_repeated_nested("b", to="nested")
# The same, but by base column name, so ndf["nested.b"] is np.repeat(ndf["b"], list_lengths)
```

It would also require adding lower-level implementations to the accessor and the extension array.

The example use-case: moving band information in ZTF DR23 catalog to light-curves, so we can merge multiple light curves with different passbands later.

**Before submitting**
Please check the following:

hombit:
It makes it possible to copy "base" column to a sub-column copying the value for each list-element

Fixes #200

### DIFF
--- a/docs/tutorials/low_level.ipynb
+++ b/docs/tutorials/low_level.ipynb
@@ -17,7 +17,12 @@
    "cell_type": "code",
    "execution_count": null,
    "id": "619f088e7ac0f327",
-   "metadata": {},
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2025-03-05T20:34:51.884860Z",
+     "start_time": "2025-03-05T20:34:51.881771Z"
+    }
+   },
    "outputs": [],
    "source": [
     "import numpy as np\n",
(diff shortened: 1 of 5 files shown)


dougbrn (changes requested):
What's here looks good to me. To fix #200, I think we need a NestedFrame level function to make this easily discoverable for the intended use case of folding base columns into nested structures. Would you be willing to write that function for this PR? If not we can merge this PR and leave the issue open. The PR suggests `add_repeated_nested` as the function, but I also like a name like `embed` for this function, whatever you think is best.

hombit:
@dougbrn 

> I think we need a NestedFrame level function

I considered the idea to add a dataframe-level function, but I decided to not doing it to keep high-level interface as slim as possible. I'm a bit afraid that we can bloat the `NestedFrame` interface with a lot of new methods which would be really hard to track. Let's keep it a part of the lower-level interface until we have more than one use-case for that.

> a name like `embed` for this function

Do you think it is worth to rename extension array and accessor methods?

dougbrn:
Per slack discussion, agree with sentiment to keep nestedframe interface lean. Suggested to make a clearer example in the docs for #200. I don't think you need to change the names for the existing functions as they fit the extension array context well as is.

hombit:
@dougbrn  I propose to solve this issue with `__setitem__`, which I implemented in a separate PR https://github.com/lincc-frameworks/nested-pandas/pull/216

PR #215, opened 2025-03-05
Move query/eval related tools to a new module

hombit:
A refactoring for better `nestedframe.core` readability


gitosaurus (approved):
LGTM

PR #216, opened 2025-03-05
Allow embedding base column with value repetition via assignment

hombit:
So now we can do this:
```python
from nested_pandas.datasets import generate_data

ndf = generate_data(5, 20)
ndf["nested.a"] = ndf.a
ndf["nested.x"] = ndf.a * ndf.b
```

But NOT this, because we check rhs to have the same index as ndf, otherwise we do assignment assuming that the value has the "flat" shape:
```python
ndf["nested.y"] = [1,2,3,4,5]  # FAILS
```

Based on PR #213

### DIFF
--- a/docs/tutorials/data_manipulation.ipynb
+++ b/docs/tutorials/data_manipulation.ipynb
@@ -17,7 +17,12 @@
   {
    "cell_type": "code",
    "execution_count": null,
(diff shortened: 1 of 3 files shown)


dougbrn (approved):
Nice, appreciate the added example!

PR #229, opened 2025-04-04
Overhaul API Reference in docs

### LINKED ISSUE #222: Document `.reduce()` returning nested data
It would be nice to document recently implemented `.reduce()` feature to return nested frames. Ideas for the documentation example case:
1. Split light curve to single band light curves, like `{"lc.band", "lc.time", "lc.flux"}` to `{"lc_r.time", "lc_r.flux", "lc_g.time", "lc_r.flux"}`.
2. Produce a periodogram `{"periodogram.freq", "periodogram.power"}` or a structure function
3. Replace light curve with a new one. For example filter data in a non-trivial way (e.g. sigma-clipping) or interpolate the data

Some of those could go to the LSDB docs instead.

dougbrn:
Our existing API reference is generated by sphinx autoapi, and generally is fairly lacking for practical usage. Some clear pain points are:

* Classes like NestedFrame are buried deep, requires clicking into the nestedframe.core module in the toc tree
* All functions for classes are displayed in one large page, making per function reference difficult
* And not autoapi's fault, but most of our functions don't have a nice example for the user to look at

This PR takes inspiration from the pandas docs, and creates a reference API that pushes the most relevant classes to the top and houses each of them in a single nicer page. Every function for those classes lives in it's own page as well. I also took the liberty here to add examples with output to our nestedframe functions as well as the count_nested utility function (which I think is not known/used nearly as much as it should be thanks to the current docs). The reduce example should close #222 as it has an example for infer_nesting.

The top-level toc tree should be more useful now, but includes the full module path, I could not figure out a way within the autosummary API/sphinx docs to modify these names unfortunately but if anyone knows how please let me know!


docs/reference/nestedframe.rst
```diff
@@ -0,0 +1,47 @@
...
+    :toctree: api/
+    
+    NestedFrame.add_nested
+    NestedFrame.from_flat
+    NestedFrame.from_lists
+
+Extended Pandas.DataFrame Interface
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. note::
```
delucchi-cmu, line 25: This looks soooooo slick in the rendered page!!

src/nested_pandas/utils/utils.py
```diff
@@ -25,6 +23,32 @@ def count_nested(df, nested, by=None, join=True) -> NestedFrame:
...
+    0  0.417022  0.184677  [{t: 8.38389, flux: 10.233443, band: 'g'}; …] ...        10
+    1  0.720324  0.372520  [{t: 13.70439, flux: 41.405599, band: 'g'}; …]...        10
+    2  0.000114  0.691121  [{t: 4.089045, flux: 69.440016, band: 'g'}; …]...        10
+    3  0.302333  0.793535  [{t: 17.562349, flux: 41.417927, band: 'g'}; …...        10
+    4  0.146756  1.077633  [{t: 0.547752, flux: 4.995346, band: 'r'}; …] ...        10
+
+    `count_nested` also allows counting by a given subcolumn, for example we
+    can count by "band" label:
+
+    >>> count_nested(nf, "nested", by="band")
```
delucchi-cmu, line 45: This is too wide for the rendered page. Best suggestion I have is to use `display.float_format` and truncate all those floating points 1-2 decimals to make the display not so wide.

image

It's nice that the display gives you a slider, but better not to need it.

delucchi-cmu (commented):
This is awesome!!

src/nested_pandas/utils/utils.py
dougbrn: Good call, I changed this to instead use "join=False" to just return a small nestedframe result instead to avoid it and teach about the join kwarg.

hombit (approved):
Looks great and thank you for doc-tests!!! Claude Code helped me with menu title issue, I'll submit it as a separate PR

PR #230, opened 2025-04-05
Add inner_dtypes to NestedDtype for sub-column dtype casting

hombit:
It is the first step for multiple nesting implementation. 

This is a draft PR, because I found the implementation to be over-complicated. Probably the better solution is using list-struct as the default data representation instead of this `inner_dtypes` system.


hombit:
This approach is way too complicated, started a new implementation at https://github.com/lincc-frameworks/nested-pandas/pull/242

PR #233, opened 2025-04-08
Remove Python path prefix from the menu docs

hombit:
My first experience with Claude Code! It was something like $1.5 of spends


dougbrn (approved):
Wow, looks great thank you Kostya (and Claude)

PR #241, opened 2025-04-17
Performance Improvement for Serialization

dougbrn:
Avoids concatenation in favor of just doing pyarrows struct array conversion from table selections.

With this, testing on my end (which is limited to a single dummy test file) indicates that there's very little if any additional memory pressure from the nested-pandas end compared to pandas and pyarrow both in full loads and partial loads. I also noticed that while pyarrow's max memory usage of partially loading a struct does not seem any smaller than a full load, so while I was surprised to not see any benefit on the nested side it looks like that is pyarrow

Here's my reproducer experiment:

```
from nested_pandas.datasets.generation import generate_data
nf = generate_data(1000,1000, seed=1)
nf.to_parquet("serial_test.parquet")
```

```
from memory_profiler import memory_usage
import pyarrow.parquet as pq

def myfunc():
  return pq.read_table("serial_test.parquet", columns=["a", "nested"]) # swap "nested" for "nested.flux"

mem = max(memory_usage(proc=myfunc))

print("Maximum memory used: {} MiB".format(mem))
```

I see very little change in the above block when swapping between partial and full loads of "nested"

### DIFF
--- a/src/nested_pandas/nestedframe/io.py
+++ b/src/nested_pandas/nestedframe/io.py
@@ -136,8 +136,16 @@ def read_parquet(
             field_names = [table.column_names[i] for i in indices]
 
             # Use iterchunks to process chunks of each column
-            chunked_arrays = [pa.concat_arrays(list(table.column(i).iterchunks())) for i in indices]
-
+            chunked_arrays = []
+            for i in indices:
+                column = table.column(i)
+                if len(column.chunks) == 1:
+                    # If there is only one chunk, use it directly
+                    # avoid copy in concat_arrays
+                    chunked_arrays.append(column.chunk(0))
+                else:
+                    # Otherwise, concatenate all chunks
+                    chunked_arrays.append(pa.concat_arrays(list(column.iterchunks())))
             structs[col] = pa.StructArray.from_arrays(
                 chunked_arrays,  # Child arrays
                 field_names,  # Field names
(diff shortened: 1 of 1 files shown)


src/nested_pandas/nestedframe/io.py
```diff
@@ -136,8 +136,16 @@ def read_parquet(
...
+            chunked_arrays = []
+            for i in indices:
+                column = table.column(i)
+                if len(column.chunks) == 1:
+                    # If there is only one chunk, use it directly
+                    # avoid copy in concat_arrays
+                    chunked_arrays.append(column.chunk(0))
+                else:
+                    # Otherwise, concatenate all chunks
+                    chunked_arrays.append(pa.concat_arrays(list(column.iterchunks())))
```
hombit, line 148: I'd prefer to never concatenate chunks, let's maybe adopt the same approach as used here?
https://github.com/lincc-frameworks/nested-pandas/blob/2f8cf4cf232632b9c8bd8ed69c0091d09b016611/src/nested_pandas/series/ext_array.py#L859

dougbrn: Gotcha, I started looking into that approach, but stumbled upon a one-liner solution that seems to do everything we want: `table.select(indices).to_struct_array()`

hombit (approved):
This is very cool

PR #242, opened 2025-04-22
Preparation for list-struct backend for NestedExtensionArray

hombit:
This PR adds a couple of helper functions to prepare a transition from struct-list chunked array to list-structure chunked array as the storage of `NestedExtensionArray` data.

### DIFF
--- a/src/nested_pandas/series/ext_array.py
+++ b/src/nested_pandas/series/ext_array.py
@@ -755,6 +755,23 @@ class NestedExtensionArray(ExtensionArray):
             return ArrowExtensionArray(self._list_array)
         return ArrowExtensionArray(self._chunked_array)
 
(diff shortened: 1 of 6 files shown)


src/nested_pandas/series/ext_array.py
```diff
@@ -755,6 +778,23 @@ def to_arrow_ext_array(self, list_struct: bool = False) -> ArrowExtensionArray:
             return ArrowExtensionArray(self._list_array)
         return ArrowExtensionArray(self._chunked_array)
 
+    def to_pyarrow_scalar(self, list_struct: bool = False) -> pa.ListScalar:
```
dougbrn, line 781: Probably a good idea to add this to the docs here: https://github.com/lincc-frameworks/nested-pandas/blob/main/docs/reference/ext_array.rst

dougbrn (approved):
Looks good, just one docs request

PR #244, opened 2025-04-23
Run CI with lowest versions of deps

hombit:
Run CI with oldest dependencies of the package while using newest versions for [dev] dependencies. It is a prototype of https://github.com/lincc-frameworks/python-project-template/issues/495 implementation

### DIFF
--- a/.github/workflows/testing-and-coverage.yml
+++ b/.github/workflows/testing-and-coverage.yml
@@ -36,3 +36,26 @@ jobs:
       uses: codecov/codecov-action@v5
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
+
+  test-lowest-versions:
+    
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.9'
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        python -m pip install --upgrade uv
+        uv venv venv
+        source venv/bin/activate
+        uv pip compile --resolution=lowest -o requirements_lowest.txt pyproject.toml
+        uv pip install --constraint=requirements_lowest.txt -e '.[dev]'
+    - name: Run unit tests with pytest
+      run: |
+        source venv/bin/activate
+        python -m pytest
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,10 +20,9 @@ requires-python = ">=3.9"
 dependencies = [
     "numpy>=2",
     # We use internal pd._libs.missing and experimental ArrowExtensionArray
-    "pandas>=2.2,<2.3",
+    "pandas>=2.2.3,<2.3",
     "pyarrow>=18",
-    "universal_pathlib"
-
+    "universal_pathlib>=0.2",
 ]
 
 [project.urls]


.github/workflows/testing-and-coverage.yml
```diff
@@ -36,3 +36,26 @@ jobs:
...
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.9'
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        python -m pip install --upgrade uv
+        uv venv venv
+        source venv/bin/activate
+        uv pip compile --resolution=lowest -o requirements_lowest.txt pyproject.toml
```
delucchi-cmu, line 56: I'm trying to understand how/why this works. 

Is it true that the `requirements_lowest` file is temporary, and not stashed anywhere? Do the lowest versions only exist in the `pyproject.toml[dependencies]`? 

Should we be populating the lowest version in the dependencies for all of our direct dependencies? I think we have, in the past, pinned versions of indirect dependencies as well.

.github/workflows/testing-and-coverage.yml
```diff
@@ -36,3 +36,26 @@ jobs:
       uses: codecov/codecov-action@v5
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
+
+  test-lowest-versions:
```
delucchi-cmu, line 40: Lemme convince myself: we don't need this workflow in the nightly smoke test because the smoke test is intended to catch issues with the latest versions of the dependencies creating problems. 

We should only see failures when we are submitting new code changes, as the pinned versions of the dependencies should not be changing and introducing new regressions against unchanged versions of our own code.

.github/workflows/testing-and-coverage.yml
```diff
@@ -36,3 +36,26 @@ jobs:
...
+      with:
+        python-version: '3.9'
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        python -m pip install --upgrade uv
+        uv venv venv
+        source venv/bin/activate
+        uv pip compile --resolution=lowest -o requirements_lowest.txt pyproject.toml
+        uv pip install --constraint=requirements_lowest.txt -e '.[dev]'
```
delucchi-cmu, line 57: You don't need `'` around dev, as this isn't running on MacOS.

delucchi-cmu (commented):
Mostly questions, since we'd likely want to explain everything well in the PPT documentation, so let's talk about it now!

.github/workflows/testing-and-coverage.yml
hombit: Just wanted to make it copy-pastable, CI is good for examples =)

.github/workflows/testing-and-coverage.yml
hombit: > Is it true that the requirements_lowest file is temporary

Yes, the file is temporary, maybe it would be better to put it to some different path. I don't think it is a good idea to add it to the repository: 1) we do need a robust way to maintain it (and PPT users), 2) there are a lot of annoying edge-cases, for example what if some version is being revoked from PyPI

> Should we be populating the lowest version in the dependencies for all of our direct dependencies? 

In the [PPT issue](https://github.com/lincc-frameworks/python-project-template/issues/495#issuecomment-2773958564) I proposed to make it a user choice, with `--resolution` you can do both: `lowest` is for all dependencies, `lowest-direct` is for direct dependencies only. I do think that we should do `lowest` in our project.

> I think we have, in the past, pinned versions of indirect dependencies as well.

I think we should do that if needed, and `--resolution=lowest` would help with that

.github/workflows/testing-and-coverage.yml
hombit: I'm also not 100% convinced about that, and I don't have any more arguments for or against adding a smoke test for that.

.github/workflows/testing-and-coverage.yml
hombit: It is actually not a macOS thing, it is zsh vs bash. I do use zsh on Linux!

delucchi-cmu (approved):

PR #246, opened 2025-04-26
Support kwargs and loading from directories in `read_parquet`

smcguire-cmu:
Passes kwargs from `read_parquet` to `pq.read_table`. Also changes how `UPath` loading works, passing the path and filesystem to read_parquet to be able to read from a directory instead of opening the path manually.


dougbrn (approved):
Nice, thank you!

PR #257, opened 2025-05-09
Change NestedExtArray.__arrow__ to return list_struct

hombit:
It is a small change which would allow us to implicitly track multiple-nested arrays: each field which is a list-struct array could be cast to a nested array.

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -1242,6 +1242,28 @@ class NestedFrame(pd.DataFrame):
 
         return results_nf
 
+    def to_pandas(self, list_struct=False) -> pd.DataFrame:
+        """Convert to an ordinal pandas DataFrame, with no NestedDtype series.
+
+        NestedDtype is cast to pd.ArrowDtype
+
+        Parameters
+        ----------
+        list_struct: bool
+            If True, cast nested columns to pandas struct-list arrow extension
+            array columns. If False (default), cast nested columns to
+            list-struct array columns.
+
+        Returns
+        -------
+        pd.DataFrame
+            Ordinal pandas DataFrame.
+        """
+        df = pd.DataFrame(self)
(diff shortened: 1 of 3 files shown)


src/nested_pandas/nestedframe/core.py
```diff
@@ -1269,13 +1291,14 @@ def to_parquet(self, path, **kwargs) -> None:
         >>> nf = generate_data(5,5, seed=1)
         >>> nf.to_parquet("nestedframe.parquet")
         """
+        df = self.to_pandas(list_struct=False)
```
dougbrn, line 1294: Not neccesarily a blocking issue, but just want to acknowledge that we're adding an additional copy operation to our parquet writing here. At scale if we run into memory issues this can be a good thing to come back to.

dougbrn (approved):
Just one performance note, but no changes requested

src/nested_pandas/nestedframe/core.py
hombit: I believe it is a shallow copy

```python
In [1]: import os
   ...: import psutil
   ...:
   ...: process = psutil.Process(os.getpid())
   ...: print(f"{process.memory_info().rss / 1024 ** 2:.2f} MiB")
175.52 MiB

In [2]: from nested_pandas.datasets import generate_data

In [3]: nf = generate_data(100_000, 100)

In [4]: print(f"{process.memory_info().rss / 1024 ** 2:.2f} MiB")
1460.98 MiB

In [5]: %time df = nf.to_pandas()
CPU times: user 527 μs, sys: 56 μs, total: 583 μs
Wall time: 516 μs

In [6]: print(f"{process.memory_info().rss / 1024 ** 2:.2f} MiB")
1461.58 MiB
```

dougbrn: Ah, wonderful thank you for verifying that!

PR #259, opened 2025-05-09
Accessor to return NestedDtype flat series

hombit:
Adding support of inner nested fields returned by `.nest["inner_nested"]` and `.nest.to_flat()`


dougbrn (approved):
Nice, thank you!

PR #260, opened 2025-05-09
Autocast list-struct columns to nested

hombit:
In `NestedFrame`, on construction or a new column assignment, check all the columns, and cast arrow list-struct columns to nested.


dougbrn (approved):
LGTM

PR #264, opened 2025-05-12
Add .nest.to_flatten_inner

hombit:
This adds a niche accessor method to "explode" an inner nested frame into the outer nested frame. It basically the same as nf.join(nf.nested.nest.to_flat()), but applied to each "higher" level nested row.

It should be useful for the cases when nested series was nested one more time with `NestedFrame.add_nested` and supposed to be flatten after that. It may be used as a part of ZTF DR22 cross-matching pipeline to concatenate multiple light-curves matched to a single object into a single multi-band light-curve.

### DIFF
--- a/src/nested_pandas/series/accessor.py
+++ b/src/nested_pandas/series/accessor.py
@@ -12,7 +12,7 @@ from numpy.typing import ArrayLike
 from pandas.api.extensions import register_series_accessor
 
 from nested_pandas.series.dtype import NestedDtype
-from nested_pandas.series.packer import pack_sorted_df_into_struct
+from nested_pandas.series.packer import pack_flat, pack_sorted_df_into_struct
 from nested_pandas.series.utils import nested_types_mapper
 
 __all__ = ["NestSeriesAccessor"]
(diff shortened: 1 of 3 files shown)


tests/nested_pandas/series/test_accessor.py
```diff
@@ -1073,3 +1073,37 @@ def test_get_list_index():
...
+def test_to_flatten_inner():
+    """Test .nest.to_flatten_inner()"""
+    nf = generate_data(10, 2)
+    nf["a"] = nf["a"].astype(pd.ArrowDtype(pa.float64()))
+    nf["b"] = nf["b"].astype(pd.ArrowDtype(pa.float64()))
+    nf = nf.assign(id=np.repeat(np.r_[0:5], 2))
+    nf = nf.rename(columns={"nested": "inner"})
+    nnf = NestedFrame.from_flat(nf, base_columns=[], on="id", name="outer")
+
+    # outer = nnf["outer"]
```
dougbrn, line 1087: remove commented out code? Or is it being kept for later?

src/nested_pandas/series/accessor.py
```diff
@@ -540,3 +540,94 @@ def clear(self) -> None:
...
+
+    def to_flatten_inner(self, field: str) -> pd.Series:
+        """Explode the nested inner field and return as a pd.Series
+
+        Works for the case of multiple nesting only, the field must represent
+        a nested series.
+
+        Each row of this Series is changed in the following way:
+        1. Each nested item in the given field is converted to a "flat" frame.
+        2. All items of other fields are repeated as many times as that frame
```
dougbrn, line 552: Is the expectation here that this will only work for doubly nested structures and not triply+ nested structures

dougbrn (approved):
Looks good just a question and a potential cleanup

tests/nested_pandas/series/test_accessor.py
hombit: Thanks!

src/nested_pandas/series/accessor.py
hombit: This works to flat the second level only, the third level will become the second level after that. Could you please suggest a docstring improvement?..

src/nested_pandas/series/accessor.py
```diff
@@ -540,3 +540,94 @@ def clear(self) -> None:
...
         raise NotImplementedError("Cannot delete fields from nested series")
+
+    def to_flatten_inner(self, field: str) -> pd.Series:
+        """Explode the nested inner field and return as a pd.Series
+
+        Works for the case of multiple nesting only, the field must represent
+        a nested series.
+
+        Each row of this Series is changed in the following way:
+        1. Each nested item in the given field is converted to a "flat" frame.
```
dougbrn, line 551:
```suggestion
        1. Each nested item in the given field is converted to a "flat" frame. If a nested item contains fields that are also nested, those are brought up as their own nested structures in the resulting "flat" frame.
```

PR #265, opened 2025-05-13
fix slow html_repr at scale

### LINKED ISSUE #262: Display of a single nested file loaded from parquet hangs indefinitely
Displaying a single nestedFrame loaded from parquet file hangs indefinitely. 

Following suggestions from this [thread](https://discovery-alliance.slack.com/archives/C04610PQW9F/p1746976462640049) I look at print statements:

```table = pq.read_table("/sdf/data/rubin/shared/lsdb_commissioning/mock_dp1/object_collection/object_lc/dataset/Norder=3/Dir=0/Npix=562.parquet")
df = table.to_pandas() 
%timeit print(df);
```
gives 
> 29 ms ± 124 μs per loop

while doing 

```from nested_pandas import NestedFrame, read_parquet
nested_df = read_parquet("/sdf/data/rubin/shared/lsdb_commissioning/mock_dp1/object_collection/object_lc/dataset/Norder=3/Dir=0/Npix=562.parquet")
%timeit print(nested_df)
```
gives
> 70.2 ms ± 342 μs per loop

But just plainly trying to run
`nested_df`
in Jupyter notebook is took 11 minutes and 10 seconds.

dougbrn:
Closes #262.

The culprit seems to be the html_repr code, as there's a lot of styling happening that currently is being applied to every row in the dataframe, even if just a small subset is displayed. This PR fixes this by truncating the html dataframe upfront, avoiding excess styling.

I verified that this codeblock reproduces the issue:
```
from nested_pandas.datasets import generate_data
nf = generate_data(270000,13,seed=1)
nf
```

This PR resolves the very slow behavior for this codeblock, and I also verify that it resolves the slowdown for the specific case in #262

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -131,9 +131,16 @@ class NestedFrame(pd.DataFrame):
                 return None
             return chunk.to_html(max_rows=1, max_cols=5, show_dimensions=True, index=False, header=False)
 
+        # Handle sizing, trim html dataframe if output will be truncated
+        df_shape = self.shape
+        if pd.get_option("display.max_rows") is None or df_shape[0] > pd.get_option("display.max_rows"):
+            html_df = self.head(pd.get_option("display.min_rows") + 1)
+        else:
+            html_df = self.copy()
+
         # replace index to ensure proper behavior for duplicate index values
-        index_values = self.index
-        html_df = self.reset_index(drop=True)
+        index_values = html_df.index
+        html_df = html_df.reset_index(drop=True)
 
         # Apply repacking to all nested columns
         repr = html_df.style.format(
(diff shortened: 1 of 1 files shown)


src/nested_pandas/nestedframe/core.py
```diff
@@ -153,14 +160,14 @@ def map_true_index(index):
         # Recover some truncation formatting, limited to head truncation
         if pd.get_option("display.max_rows") is None:
             return repr.to_html(max_rows=0)
-        elif repr.data.shape[0] > pd.get_option("display.max_rows"):
+        elif df_shape[0] > pd.get_option("display.max_rows"):
```
gitosaurus, line 163: I think I understand this change.  Is the idea to avoid creating a monster HTML representation *first* and then clipping it?  Or is it about having an accurate count of the original shape (`df_shape`) for the test?  (Or both, actually, is what it looks like.  I believe I understand now.)

gitosaurus (approved):
Looks good!  The logic is compact enough that maybe it could use a couple of comments to explain motivation.

src/nested_pandas/nestedframe/core.py
dougbrn: Right, the change to df_shape here is basically what you describe, but more specifically because repr.data may now be truncated, we want to use the original shape information to determine the truncation setting to display (basically to preserve the "..." that you see at the end of a truncated dataframe).

Issue #270, opened 2025-05-20
Use NestedFrame for element representation


hombit:
**Feature request**

Use `NestedFrame` instead of `pd.DataFrame` when accessing a single nested element, e.g. with `nested_series.iloc[0]`. We also need a benchmark for that, because pandas data-frame construction is already a bottle-neck and we wouldn't like to make it much worse.

**Before submitting**
Please check the following:

- [x] I have described the purpose of the suggested change, specifying what I need the enhancement to accomplish, i.e. what problem it solves.
- [ ] I have included any relevant links, screenshots, environment information, and data relevant to implementing the requested feature, as well as pseudocode for how I want to access the new functionality.
- [ ] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

PR #271, opened 2025-05-20
Propagate inner nested columns to elements

hombit:
Use inner nested dtypes for "element" data-frames. I think we should also use `NestedFrame` for the output, but I'd like to have a benchmark for that, so I opened https://github.com/lincc-frameworks/nested-pandas/issues/270


dougbrn (approved):
LGTM!

dougbrn (approved):
Looks good with new changes

PR #281, opened 2025-06-06
Add `from_pyarrow` method to load nested frame from pyarrow table

smcguire-cmu:
Refactors the `read_parquet` method to allow for loading a nested frame from a pyarrow table.


dougbrn (approved):
Nice, thanks @smcguire-cmu !

PR #289, opened 2025-06-16
Autocast lists option for read_parquet

### LINKED ISSUE #268: Autocast List Columns?
**Feature request**
Have been playing with this idea for the last day or so, where essentially; if we think that nested columns are a more supported interface than lists, why not just cast them directly?

The main case I'm thinking of is on read, with the option to turn it off via a kwarg (`cast_lists`). The obvious issue being that if there are multiple lists that are part of the same structure, our casting would produce multiple separate nested columns, but it may be easier from an API perspective for users to combine those nested columns into a single column (potentially using a new function called `combine_nested`) rather than using the nest_lists interface, e.g.
```
import nested_pandas as npd
nf = npd.read_parquet("input.parquet") # has list-columns; flux, mjd, band

# nf has 3 nested structures; flux (with sub-column flux), mjd (with subcolumn mjd), and band (with subcolumn band)

# create a single nested structure
nf = nf.combine_nested(["flux", "mjd","band"], name="lc")
# now nf has 1 nested structure; lc (with sub-columns flux, mjd, and band)
```

The strength of this is that it's more clear to users that flux, mjd and band are related because the nested representation will provide the dimensionality of the lists, so you can more easily see what fields are equal length. With lists, you need to either go off name context or check the lengths yourself.

Another strength is that these columns are usable right away in the API for filtering, counting, reduce, etc. You could even avoid the combination step if you just wanted to use the columns for reduce right after loading.

The downsides I see are that there's some overhead to casting this, and generating the offsets with potential duplication of offsets in cases like above.

**Before submitting**
Please check the following:

dougbrn:
Closes #268

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -7,6 +7,7 @@ import numpy as np
 import pandas as pd
 import pyarrow as pa
 from nested_pandas import NestedDtype, NestedFrame, datasets
+from nested_pandas.utils import count_nested
 
 
 class AssignSingleDfToNestedSeries:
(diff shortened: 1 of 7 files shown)


dougbrn:
merge/rebase mucked this up, opening fresh pr

PR #307, opened 2025-07-11
added min function

Graciaaa3:
## Change Description

Added the min() function extension to work for nested structure. 

## Solution Description

The inherited min() function from pandas will throw an error for nested structure. The min() function for nested-pandas extends that by allowing client to indicate whether to include the nested columns or not with an extra parameter. The min() extension only supports column-wise minimum with a default to skip missing values. 

Still need to finish the example part for docstring and adding unit tests to test the implementation.

## Code Quality
- [ ] My code builds (or compiles) cleanly without any errors or warnings

### New Feature Checklist

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -63,7 +63,9 @@ class NestedFrame(pd.DataFrame):
             if not isinstance(dtype, pd.ArrowDtype):
                 continue
             pa_type = dtype.pyarrow_dtype
-            if not is_pa_type_a_list(pa_type) and not (struct_list and pa.types.is_struct(pa_type)):
+            if not is_pa_type_a_list(pa_type) and not (
+                struct_list and pa.types.is_struct(pa_type)
+            ):
                 continue
             try:
                 nested_dtype = NestedDtype(pa_type)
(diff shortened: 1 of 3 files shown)


src/nested_pandas/nestedframe/core.py
```diff
@@ -673,6 +673,71 @@ def drop(
...
+        # handle the nested columns:
+        # collect mins for all nested columns as a list, each is a series
+        nested_mins = []
+        for nest_col in self.nested_columns:
+            nested_df = self[nest_col].nest.to_flat()  # returns a DataFrame
+
+            # prefix column to indicate source
+            nested_df.columns = [f"{nest_col}." + col for col in nested_df.columns]
+
+            # compute min on falt DataFrame
```
dougbrn, line 732: very minor: falt -> flat

src/nested_pandas/nestedframe/core.py
```diff
@@ -673,6 +673,71 @@ def drop(
...
+
+        Returns
+        -------
+        pd.Series 
+
+        Examples
+        --------
+
+        """
+
```
dougbrn, line 713: up front, I think it would be good to check if there are no nested columns, in which case, just return the super().min(**kwargs). That way we can more reliably have min behave exactly like pandas min for "normal" dataframes.

dougbrn (commented):
This is looking good so far @Graciaaa3 ! Just left two minor comments, and yeah filling out the examples section and writing some tests are good next steps

src/nested_pandas/nestedframe/core.py
```diff
@@ -673,6 +673,79 @@ def drop(
...
+        # if no nested columns, directly call super min
+        if not self.nested_columns:
+            return super().min(numeric_only=numeric_only, **kwargs)
+    
+        # handle base columns
+        base_col = [col for col in self.columns if col not in self.nested_columns]
+        # need to check that base column is not empty???
```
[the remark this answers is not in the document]
Graciaaa3: resolved

src/nested_pandas/nestedframe/core.py
```diff
@@ -693,6 +707,77 @@ def drop(
...
+        a              0.000114
+        b              0.184677
+        nested.t       0.547752
+        nested.flux    1.828828
+        nested.band           g
+        dtype: object
+
+        """
+
+        # if no nested columns, directly call super min
```
gitosaurus, line 757: This comment is not necessary, since the code that follows it is perfectly clear.

PR #311, opened 2025-07-18
Use intersphinx to be able to directly reference Pandas methods.

gitosaurus:
Also fix some misspoken docstrings as an example of this new capability.

## Change Description

Enable the ability to reference Pandas methods from nested-pandas docstrings so that they render in the final HTML documentation as legible, clickable links.

## Code Quality

### New Feature Checklist
- [X] I have added or updated the docstrings associated with my feature using the [NumPy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html)

### Documentation Change Checklist
- [X] Any updated docstrings use the [NumPy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -61,7 +61,6 @@ autosummary_imported_members = True
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
     "pandas": ("https://pandas.pydata.org/pandas-docs/stable/", None),
-    "pyarrow": ("https://arrow.apache.org/docs/python", None),
     "numpy": ("https://numpy.org/doc/stable/", None),  # optional but handy
 }
 


docs/conf.py
```diff
@@ -52,6 +58,12 @@
 # Customize display of autosummary entries
 autosummary_imported_members = True
 
+intersphinx_mapping = {
+    "python": ("https://docs.python.org/3", None),
+    "pandas": ("https://pandas.pydata.org/pandas-docs/stable/", None),
+    "numpy": ("https://numpy.org/doc/stable/", None),  # optional but handy
```
dougbrn, line 64: Not sure if we have any links, but adding pyarrow here might be good

Graciaaa3 (commented):
Documentation is making much more sense to me! intersphinx is amazing :)

docs/conf.py
gitosaurus: Great idea!  Will do.

gitosaurus: Unfortunately, I had to revert the change, since pyarrow doesn't have an intersphinx inventory like the other packages.

dougbrn (approved):
Looks good! The rtd fail does look like the url issue with sdss spectra, so hopefully that builds properly when merged

PR #322, opened 2025-08-05
adding describe

Graciaaa3:
## Change Description

Added the describe() function extension to work for nested structure.

## Solution Description

Extending the describe function to work for nested frame. Apply the `super().describe()` on base columns and handles the nested columns separately by applying `.nest.to_flat()` to each column and calls describe() on the resulting flat DataFrame. Column names are prefixed with <nested_col>. for clarity. The basic behavior is similar to pandas' describe, with an extra `exclude_nest=`to return only base column stats if set to True.

Because each part is handled separately, if certain columns don't match the include/exclude criteria (e.g., object), they are skipped instead of raising an error. If no columns match the criteria, a combined ValueError is raised listing the errors encountered.

Due to the way nested columns store data with PyArrow, the strings in nested columns are not included in object datatype as they are not being casted. 

## Code Quality

### New Feature Checklist
- [x] I have updated the tutorial to highlight my new feature (if appropriate)
- [x] I have added unit/End-to-End (E2E) test cases to cover my new feature

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -917,7 +917,7 @@ class NestedFrame(pd.DataFrame):
 
         result = []
         errors = []
-        check = ["Base_columns"]  # a list of all possible columns to call describe()
+        check = ["_base"]  # a list of all possible columns to call describe()
         if not exclude_nest:
             check.extend(self.nested_columns)
 
@@ -926,7 +926,7 @@ class NestedFrame(pd.DataFrame):
 
         for checkable in check:
             # check the base columns
-            if checkable == "Base_columns":
+            if checkable == "_base":
                 try:
                     base_col = [col for col in self.columns if col not in self.nested_columns]
                     base_desc = (


src/nested_pandas/nestedframe/core.py
```diff
@@ -858,6 +858,111 @@ def max(self, exclude_nest: bool = False, numeric_only: bool = False, **kwargs):
...
+
+        See Also
+        --------
+        :meth:`pandas.DataFrame.describe`
+
+        """
+
+        result = []
+        errors = []
+        check = ["Base_coloumns"]   # a list of all possible columns to call describe()
```
gitosaurus, line 919:
```suggestion
        check = ["Base_columns"]   # a list of all possible columns to call describe()
```

src/nested_pandas/nestedframe/core.py
```diff
@@ -858,6 +858,111 @@ def max(self, exclude_nest: bool = False, numeric_only: bool = False, **kwargs):
...
+        if not exclude_nest:
+            check.extend(self.nested_columns)
+
+        if not self.nested_columns:
+            return NestedFrame(super().describe(percentiles=percentiles, include=include, exclude=exclude))
+        else:
```
gitosaurus, line 925: Since you have an early return (which is good!) you don't need to put the rest of the function under an `else:`.  A pattern of handling edge conditions with a series of early returns makes the code more readable.  After all the edges are handled, then the function's main work takes place.

src/nested_pandas/nestedframe/core.py
```diff
@@ -858,6 +858,111 @@ def max(self, exclude_nest: bool = False, numeric_only: bool = False, **kwargs):
...
+            for checkable in check:
+                # check the base columns
+                if checkable == "Base_coloumns":
```
gitosaurus, line 928:
```suggestion
                if checkable == "Base_columns":
```

Since this is a piece of internal state, you can use a simple identifier like `"base"`.  It is also good practice to use a enum or constant for these things, so that spelling inconsistencies are caught by the linter.

src/nested_pandas/nestedframe/core.py
```diff
@@ -858,6 +858,111 @@ def max(self, exclude_nest: bool = False, numeric_only: bool = False, **kwargs):
...
+                    except ValueError as err:
+                        # continue if value error caused by no matching type for nested columns
+                        errors.append(f"Nested column '{checkable}': {err}")
+                        continue
+
+                    result.append(nested_desc)
+            
+        # if no result added, raise combined value error with all the error messages
+        if not result:
+            raise ValueError(f"All columns in {check} failed.\n" + "\n".join(errors))
```
gitosaurus, line 962: This reads really well, and it's quite clear what's happening! 😃

dougbrn (changes requested):
Looks great, just one request!

src/nested_pandas/nestedframe/core.py
Graciaaa3: Got it! I have deleted the `else` branch here.

src/nested_pandas/nestedframe/core.py
Graciaaa3: Corrected the typo

dougbrn (approved):
Nice, looks good. Great work @Graciaaa3 !

src/nested_pandas/nestedframe/core.py
```diff
@@ -858,6 +858,114 @@ def max(self, exclude_nest: bool = False, numeric_only: bool = False, **kwargs):
...
+
+        Generate descriptive statistics, including nested columns with prefix to indicate the source.
+
+        Descriptive statistics include those that summarize the central tendency,
+        dispersion and shape of a dataset's distribution, excluding NaN values,
+        similar to the behavior of `pandas.DataFrame.describe()`.
+
+        Nested columns use `pyarrow` data types for efficiency, which are not always
+        directly compatible with pandas' type-based filtering.
+        - pyarrow strings are not viewed as object type.
```
dougbrn, line 872: Minor Nit: I just gave an approval, but realized by viewing the [readthedocs build](https://nested-pandas--322.org.readthedocs.build/en/322/reference/api/nested_pandas.NestedFrame.describe.html) that the last two sentences are separated properly because you need a newline separating them in the docstring itself. Would be good to fix, but also very very minor

src/nested_pandas/nestedframe/core.py
```diff
@@ -858,6 +858,114 @@ def max(self, exclude_nest: bool = False, numeric_only: bool = False, **kwargs):
...
+        - numerical types from pyarrow (i.e., int, double) are still matched by pandas'
+          `np.number`, so filtering with `include=[np.number]` will include numeric nested columns.
```
dougbrn, line 874: Ah, and this is rendering as a callout box. Okay probably worth fixing these, just take a look at the readthedocs build link above and fix the markdown issues and this should be good.

Graciaaa3: Noted! I have made the change

gitosaurus (approved):
Looks good!

Issue #284, opened 2025-06-13
Cannot use Series methods with nested columns


hombit:
**Bug report**

Pandas' `.query()` and `.eval()` support some methods calls on columns, for example `df.query("a.isna()")` or `df.eval("b.fillna(33.33)")`. However, this doesn't work with nested columns:

```python
from nested_pandas.datasets import generate_data

nf = generate_data(12, 5)
nf = nf.query("a.notna()")  # works
_ = nf["nested"].notna()  # works
_ = nf.query("nested.notna()")  # fails
```

```
File ~/.virtualenvs/v/lib/python3.13/site-packages/pandas/core/computation/expr.py:413, in BaseExprVisitor.visit(self, node, **kwargs)
    411 method = f"visit_{type(node).__name__}"
    412 visitor = getattr(self, method)
--> 413 return visitor(node, **kwargs)

File ~/.virtualenvs/v/lib/python3.13/site-packages/pandas/core/computation/expr.py:659, in BaseExprVisitor.visit_Call(self, node, side, **kwargs)
    657 def visit_Call(self, node, side=None, **kwargs):
    658     if isinstance(node.func, ast.Attribute) and node.func.attr != "__call__":
--> 659         res = self.visit_Attribute(node.func)
    660     elif not isinstance(node.func, ast.Name):
    661         raise TypeError("Only named functions are supported")

File ~/.virtualenvs/v/lib/python3.13/site-packages/pandas/core/computation/expr.py:646, in BaseExprVisitor.visit_Attribute(self, node, **kwargs)
    644 resolved = self.visit(value).value
    645 try:
--> 646     v = getattr(resolved, attr)
    647     name = self.env.add_tmp(v)
    648     return self.term_type(name, self.env)

File ~/.virtualenvs/v/lib/python3.13/site-packages/nested_pandas/nestedframe/expr.py:168, in _NestedFieldResolver.__getattr__(self, item_name)
    166     result.flat_nest = self._flat_nest
    167     return result
--> 168 raise AttributeError(f"No attribute {item_name}")

AttributeError: No attribute isna
```

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.
- [ ] If I have a solution in mind, I have provided an explanation and/or pseudocode and/or task list.

PR #331, opened 2025-08-18
NestedSeries Implementation

### LINKED ISSUE #304: Implement `NestedSeries`
**Feature request**

Currently we extend pandas interface in three recommended ways: 1) `NestedDtype`, 2) `NestedExtensionArray`, and 3) `NestedAccessor` (aka `.nest`). However we found that these classes are not very convenient for the end-user, and that lead us to `NestedFrame`, which inherits `pd.DataFrame`.

Now we also see that we would highly benefit from exposing more API directly through series, see e.g. #274. Here I propose to implement `NestedSeries` and rethink the interface of both `NestedSeries` and `NestedAccessor` for better usability.

**Before submitting**
Please check the following:

dougbrn:
Resolves #304. This PR has gotten large enough that it might have been better to split it up into a few smaller steps, sorry about that. I've yet to write documentation, but I think it makes sense to write it as a follow up PR while we iron out any implementation/api behaviors here. 

Some design choices I made here, which we can definitely do differently:
* NestedSeries takes on a set of the nest accessor functions for direct use
* NestedSeries works with non-nested dtypes just as a normal pandas series, but nested specific properties/methods are tagged with a decorator which will throw an exception for attempted use with a non-nested dtype
* When returning a non-nested series, still try to return a native pandas series. Don't use NestedSeries as an everywhere replacement for pandas series when it's not needed.
* For masking, return result as a NestedSeries always, instead of sometimes as a NestedFrame

### DIFF
--- a/src/nested_pandas/series/accessor.py
+++ b/src/nested_pandas/series/accessor.py
@@ -436,22 +436,16 @@ class NestSeriesAccessor(Mapping):
 
         flat_chunked_array = pa.chunked_array(flat_chunks, type=self._series.dtype.fields[field])
 
-        if isinstance(self._series.dtype.field_dtype(field), NestedDtype):
-            # If the field is a nested dtype, return as NestedSeries
-            return NestedSeries(
-                flat_chunked_array,
-                dtype=self._series.dtype.field_dtype(field),
-                index=self.get_flat_index(),
-                name=field,
-                copy=False,
-            )
-        return pd.Series(
+        flat_series = pd.Series(
             flat_chunked_array,
             dtype=self._series.dtype.field_dtype(field),
             index=self.get_flat_index(),
             name=field,
             copy=False,
         )
+        if isinstance(self._series.dtype.field_dtype(field), NestedDtype):
+            return NestedSeries(flat_series, copy=False)
+        return flat_series
 
     def get_list_series(self, field: str) -> pd.Series:
         """Get the list-array field as a Series


src/nested_pandas/series/nestedseries.py
```diff
@@ -0,0 +1,89 @@
...
+
+class NestedSeries(pd.Series):
+    """
+    A Series that can contain nested data structures, such as lists or dictionaries.
+    This class extends the functionality of a standard pandas Series to handle nested data.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
```
gitosaurus, line 27: What happens if a user does binary operations on a `NestedSeries` with a `Series`?  I suspect that you may want to follow the same procedure when extending a Pandas class that `_SeriesFromNest` does, [here](https://github.com/lincc-frameworks/nested-pandas/blob/919fe8233f59c19e545bb01ba4cad372a2e0a23c/src/nested_pandas/nestedframe/expr.py#L52-L74).

I had been wondering whether `_SeriesFromNest` and `NestedSeries` could be dovetailed, but on reflection I do think they are serving different purposes: the former tracks a series (field) extracted from a nest, and the latter represents the nest as a first-class object. Do you agree?

I wonder if this means that this PR resolves (or helps resolve) #284.

src/nested_pandas/series/accessor.py
```diff
@@ -585,17 +606,18 @@ def to_flatten_inner(self, field: str) -> pd.Series:
         >>> from nested_pandas import NestedFrame
         >>> from nested_pandas.datasets import generate_data
         >>> nf = generate_data(5, 2, seed=1).rename(columns={"nested": "inner"})
+        >>> nf["b"] = "b"  # Shorten width of example output
```
gitosaurus, line 609: Funny! 🙏 for the comment. Is that because 'black' formatting interferes with doctests?

src/nested_pandas/series/accessor.py
```diff
@@ -479,20 +490,30 @@ def get_list_series(self, field: str) -> pd.Series:
...
 
-        # If the key is a single string, return the flat series for that field
+        # A list of fields may return a pd.Series or a NestedSeries depending
+        # on the number of fields requested and their dtypes
         if isinstance(key, list):
             new_array = self._series.array.view_fields(key)
-            return pd.Series(new_array, index=self._series.index, name=self._series.name)
+            # if len(key) == 1 and not isinstance(new_array.dtype.field_dtype(key[0]), NestedDtype):
+            #    # If only one field is requested, return it as a pd.Series
+            #    return self._series[key[0]]
```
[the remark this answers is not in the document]
dougbrn: Dead, good catch

src/nested_pandas/series/nestedseries.py
dougbrn: By binary operations, do you mean something like `nf["nested"] > nf["c"]`? I'm interested in being able to handle the use case you're describing, but I can't think of an example operation to support, as this example doesn't really make logical sense for a nested column.

For `_SeriesFromNest` and `NestedSeries`, that distinction of purposes makes sense to me. I do wonder if the NestedSeries implementation should be using `_SeriesFromNest` in any capacity as theres several instances where `Series` are extracted from `NestedSeries`, maybe it already does if `_SeriesFromNest` is embedded more deeply in the codebase, fully admit I'm not that familiar with it.

I think it takes a step towards, #284, perhaps. But I would imagine full resolution of #284 requires some work on NestedFrame as well

src/nested_pandas/series/nestedseries.py
```diff
@@ -0,0 +1,89 @@
...
+        result = func(*args, **kwargs)
+        return result
+
+    return wrapper
+
+
+class NestedSeries(pd.Series):
+    """
+    A Series that can contain nested data structures, such as lists or dictionaries.
```
hombit, line 21:
```suggestion
    A Series that can contain nested data structures, such as lists or data-frames.
```

src/nested_pandas/series/accessor.py
```diff
@@ -434,6 +436,15 @@ def get_flat_series(self, field: str) -> pd.Series:
...
 
+        if isinstance(self._series.dtype.field_dtype(field), NestedDtype):
+            # If the field is a nested dtype, return as NestedSeries
+            return NestedSeries(
+                flat_chunked_array,
+                dtype=self._series.dtype.field_dtype(field),
+                index=self.get_flat_index(),
+                name=field,
+                copy=False,
+            )
```
hombit, line 447: Opinionated, but I'd do:

```python
        flat_series = pd.Series(
            flat_chunked_array,
            dtype=self._series.dtype.field_dtype(field),
            index=self.get_flat_index(),
            name=field,
            copy=False,
        )
        if isinstance(self._series.dtype.field_dtype(field), NestedDtype):
            return NestedSeries(flat_series, copy=False)
        return flat_series
```

hombit (approved):
Looks great, thank you!

src/nested_pandas/series/nestedseries.py
```diff
@@ -0,0 +1,89 @@
+import pandas as pd
+
+from nested_pandas.series.dtype import NestedDtype
+
+
+def nested_only(func):
```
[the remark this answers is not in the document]
dougbrn: I don't think so, `nested_only` is positioned to evaluate a function whereas `_check_series` directly operates on the series and has a superior error message to the one that nested_only would generate for that use case, imo.

gitosaurus (approved):
LGTM!

PR #327, opened 2025-08-14
adding explode

Graciaaa3:
## Change Description

Added the `explode()` function extension to work for nested structure. In addition to regular explode on the base columns, the extension includes "unnesting" single nested column by expanding it.

- [ ] My PR includes a link to the issue that I am addressing

## Solution Description

Extending the `explode` function to work for nested frame. 

Currently, the function only support single nested column explosion. If the given `column` includes multiple nested columns or includes a mix of nested column and base column, an `ValueError` will be raised. The exploded nest columns will be added to the right of the rest of the frame with a join. Nested column names are prefixed with <nested_col>. for clarity. 

Apply the super().explode() on base columns for regular pandas' explode.

## Code Quality
- [x] I have read the Contribution Guide
- [x] My code follows the code style of this project
- [x] My code builds (or compiles) cleanly without any errors or warnings
- [x] My code contains relevant comments and necessary documentation

### New Feature Checklist
- [x] I have added or updated the docstrings associated with my feature using the [NumPy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html)
- [x] I have updated the tutorial to highlight my new feature (if appropriate)
- [x] I have added unit/End-to-End (E2E) test cases to cover my new feature
- [ ] My change includes a breaking change
  - [ ] My change includes backwards compatibility and deprecation warnings (if possible)

PR #332, opened 2025-08-19
Revisit `nf.explode()` for multiple columns and duplicated index

hombit:
I’ve actually been working on my own implementation of `NestedFrame.explode` in parallel with #327. Here I merge the two implementations but keep the test suites separate, since they cover different cases


Graciaaa3 (approved):
I think it looks great!

PR #337, opened 2025-08-25
fix nest_lists for non-unique index

### LINKED ISSUE #334: Identical catalog indices cause nest_lists to duplicate rows
**Bug report**
I'm using ZTF DR22, which has some duplicate indices (_healpix_29). One object has observations in two different bands, which might have the same _healpix_29 indices. When this occurs, the nest_lists function nests the list into a copy of each pre-existing row, yielding a duplicate of each one. See the attached screenshot, in which this object starts out with 2 rows and has 4 after nest_lists. Also notice the mismatch between nepochs and the length of the nested lists indicating which row is "fake".

Image

Here is the code snippet in the screenshot:
`
cone = lsdb.ConeSearch(ra=285, dec=-15, radius_arcsec=1*3600)
ztf_22_scattered = lsdb.open_catalog('/epyc/data3/hats/catalogs/ztf_dr22/ztf_lc', margin_cache='https://data.lsdb.io/hats/ztf_dr22/ztf_lc_10arcs', search_filter = cone)
ztf_22 = ztf_22_scattered.nest_lists(list_columns=["hmjd", "mag", "magerr", "catflags", "clrcoeff"], name="lc")
ztf = ztf_22.compute()
ztf.loc[2050364164511734616]
`

**Before submitting**
Please check the following:

- [✓ ] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [ ✓] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.

dougbrn:
Fixes #334

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -596,7 +596,7 @@ class NestedFrame(pd.DataFrame):
 
         # join the nested column to the base_column df
         if base_columns is not None:
-            return df[base_columns].join(packed_df)
+            return pd.concat([df[base_columns], packed_df], axis=1)
         # or just return the packed_df as a nestedframe if no base cols
         else:
             return NestedFrame(packed_df.to_frame())
--- a/tests/nested_pandas/nestedframe/test_nestedframe.py
+++ b/tests/nested_pandas/nestedframe/test_nestedframe.py
@@ -2032,6 +2032,17 @@ def test_nest_lists():
         res = ndf.nest_lists("nested", ["c", "b"])
 
 
+def test_nestlists_nonunique_index():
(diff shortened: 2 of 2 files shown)


src/nested_pandas/nestedframe/core.py
```diff
@@ -596,7 +596,7 @@ def from_lists(cls, df, base_columns=None, list_columns=None, name="nested"):
 
         # join the nested column to the base_column df
```
gitosaurus, line 597:
```suggestion
        # concat the nested column to the base_column df
```
(or even simply delete the comment)

gitosaurus (approved):
Looks good!

PR #349, opened 2025-09-04
Doc Tweaks for `NestedFrame.to_pandas`

### LINKED ISSUE #336: Add `NestedFrame.to_pandas` for NestedFrame->DataFrame
Realized it we didn't have a ticket for this, but have heard a few times that users would like a `NestedFrame.to_pandas` option. Even though `pd.DataFrame(NestedFrame)` is a nice option here, I do think it would be nice to have `to_pandas` as an intuitive alternative.

dougbrn:
Closes #336

### DIFF
--- a/docs/reference/nestedframe.rst
+++ b/docs/reference/nestedframe.rst
@@ -51,4 +51,5 @@ I/O
     :toctree: api/
 
     NestedFrame.to_parquet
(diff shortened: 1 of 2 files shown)


gitosaurus (approved):
LGTM!

Issue #355, opened 2025-09-11
Renaming for v0.6.0


dougbrn:
**NestedFrame**
NestedFrame.add_nested -> join_nested

**Nested Series**
fields -> columns
to_flat([fields]) -> explode

**Nested Accessor**
fields -> columns
with_field -> set_column
with_list_field -> set_list_column
with_flat_field -> set_flat_column
with_filled_field -> set_filled_column
without_field -> drop(column=)
get_flat_series(field) x remove
get_list_series(fieild) x remove
query_flat(expr) -> query
get_flat_index() -> @property flat_index

**NestedDtype**
@property fields -> column_dtypes, change to be Series
@property names -> columns
field_dtype(field) x remove
@property field_dtypes x remove
@property field_names x remove
@property struct_list_pa_dtype -> _struct_list_pa_dtype
@property pyarrow_dtype -> pyarrow_dtype
from_fields -> from_columns

All functions that have changed names should have a deprecation warning for the old name to ensure compatibility for a time

PR #360, opened 2025-09-16
API Renaming

dougbrn:
For #355, will resolve after full merge of v0.6 branch. Because this is a PR not to main the CI outside of readthedocs didn't engage, but can verify that the tests run locally, and there will also be a review to merge v0_6 into main.

### DIFF
--- a/docs/pre_executed/nested_spectra.ipynb
+++ b/docs/pre_executed/nested_spectra.ipynb
@@ -280,7 +280,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 8,
+   "execution_count": null,
    "metadata": {},
    "outputs": [
     {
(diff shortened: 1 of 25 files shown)


hombit (commented):
Great job, Doug! Few questions:

- Are we deprecating `with_*` methods?
- In the warnings, should we say that it would be dropped in v0.7.0?
- Let's maybe use [`deprecated` package](https://pypi.org/project/Deprecated/), I think IDEs and linters recognize it and help people to find the usage of deprecated code in their codebase

dougbrn:
@hombit

Yeah each of the `with_*` methods has been deprecated within the accessor interface, generally replaced "with" with "add" and "field" with "column" for those method names. I did have a thought that "set" might be better to use than "add", since it can overwrite existing columns in some cases, but not sure if you have an opinion on that. #355 should be up to date with the various renames

The deprecated package is a good idea, and I can signal in those warnings that they'll be removed in v0.7.0

hombit:
@dougbrn I'm fine with `set_*`

hombit (approved):
Looks great! Unfortunately I cannot see deprecation warnings yet, but I guess it is because we are not on v0.6 yet...

PR #403, opened 2025-11-04
Raise a helpful error for failed partial loading of list-struct columns

### LINKED ISSUE #394: Improve handling of subcolumn selection for list-struct columns
**Bug report**

Nested-Pandas (and Pandas) doesn't support subcolumn selection for list-struct columns, because `pyarrow` also doesn't support it:
https://github.com/apache/arrow/issues/46329

However, when list-struct column can be loaded and used with LSDB/nested-pandas, and from LSDB/nested-pandas user's perspective it looks exactly like struct-list column. However when user attempts to select subcolumns from list-struct column it fails a non-helpful error message. I believe we should at least give more human-readable error, and at most allow sub-column loading via loading the whole nested column and selecting subcolumns in-memory.

Reproducible code:

```python
import lsdb
import nested_pandas as npd


# Get some data, save it with pyarrow, so we are getting list-struct column
nf = lsdb.open_catalog('s3://ipac-irsa-ztf/contributed/dr23/lc/hats').head()
table = pa.Table.from_pandas(nf)
pa.parquet.write_table(table, "/tmp/tmp.parquet")

# Try to get a sub-column
npd.read_parquet('/tmp/tmp.parquet', columns=['lightcurve.hmjd'])
```

<details><summary>Traceback</summary>

```
---------------------------------------------------------------------------
ArrowInvalid                              Traceback (most recent call last)
Cell In[1], line 11
      8 pa.parquet.write_table(table, "/tmp/tmp.parquet")
     10 # Try to get a sub-column
---> 11 npd.read_parquet('/tmp/tmp.parquet', columns=['lightcurve.hmjd'])

File ~/.virtualenvs/lsdb-release/lib/python3.13/site-packages/nested_pandas/nestedframe/io.py:103, in read_parquet(data, columns, reject_nesting, autocast_list, **kwargs)
    100 # Otherwise convert with a special function
    101 else:
    102     data, filesystem = _transform_read_parquet_data_arg(data)
--> 103     table = pq.read_table(data, filesystem=filesystem, columns=columns, **kwargs)
    105 # Resolve partial loading of nested structures
    106 # Using pyarrow to avoid naming conflicts from partial loading ("flux" vs "lc.flux")
    107 # Use input column names and the table column names to determine if a column
    108 # was from a nested column.
    109 if columns is not None:

File ~/.virtualenvs/lsdb-release/lib/python3.13/site-packages/pyarrow/parquet/core.py:1899, in read_table(source, columns, use_threads, schema, use_pandas_metadata, read_dictionary, binary_type, list_type, memory_map, buffer_size, partitioning, filesystem, filters, ignore_prefixes, pre_buffer, coerce_int96_timestamp_unit, decryption_properties, thrift_string_size_limit, thrift_container_size_limit, page_checksum_verification, arrow_extensions_enabled)
   1885     # TODO test that source is not a directory or a list
   1886     dataset = ParquetFile(
   1887         source, read_dictionary=read_dictionary,
   1888         binary_type=binary_type,
   (...)   1896         page_checksum_verification=page_checksum_verification,
   1897     )
-> 1899 return dataset.read(columns=columns, use_threads=use_threads,
   1900                     use_pandas_metadata=use_pandas_metadata)

File ~/.virtualenvs/lsdb-release/lib/python3.13/site-packages/pyarrow/parquet/core.py:1538, in ParquetDataset.read(self, columns, use_threads, use_pandas_metadata)
   1530         index_columns = [
   1531             col for col in _get_pandas_index_columns(metadata)
   1532             if not isinstance(col, dict)
   1533         ]
   1534         columns = (
   1535             list(columns) + list(set(index_columns) - set(columns))
   1536         )
-> 1538 table = self._dataset.to_table(
   1539     columns=columns, filter=self._filter_expression,
   1540     use_threads=use_threads
   1541 )
   1543 # if use_pandas_metadata, restore the pandas metadata (which gets
   1544 # lost if doing a specific `columns` selection in to_table)
   1545 if use_pandas_metadata:

File ~/.virtualenvs/lsdb-release/lib/python3.13/site-packages/pyarrow/_dataset.pyx:579, in pyarrow._dataset.Dataset.to_table()

File ~/.virtualenvs/lsdb-release/lib/python3.13/site-packages/pyarrow/_dataset.pyx:415, in pyarrow._dataset.Dataset.scanner()

File ~/.virtualenvs/lsdb-release/lib/python3.13/site-packages/pyarrow/_dataset.pyx:3704, in pyarrow._dataset.Scanner.from_dataset()

File ~/.virtualenvs/lsdb-release/lib/python3.13/site-packages/pyarrow/_dataset.pyx:3617, in pyarrow._dataset.Scanner._make_scan_options()

File ~/.virtualenvs/lsdb-release/lib/python3.13/site-packages/pyarrow/_dataset.pyx:3567, in pyarrow._dataset._populate_builder()

File ~/.virtualenvs/lsdb-release/lib/python3.13/site-packages/pyarrow/error.pxi:92, in pyarrow.lib.check_status()

ArrowInvalid: No match for FieldRef.Nested(FieldRef.Name(lightcurve) FieldRef.Name(hmjd)) in objectid: int64
filterid: int8
objra: float
objdec: float
lightcurve: list<element: struct<hmjd: double, mag: float, magerr: float, clrcoeff: float, catflags: int32>>
_healpix_29: int64
__fragment_index: int32
__batch_index: int32
__last_in_fragment: bool
__filename: string
```

</details>

Originated from https://github.com/astronomy-commons/hats-import/issues/591#issuecomment-3452897831

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, information about my environment, and any applicable data others will need to reproduce the problem.
- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.

dougbrn:
Resolves #394 using the lowest-effort approach of checking and throwing an error. I decided not to do the higher-effort approach of handling this dynamically, because there isn't a more performant way to do it than doing a full load and then performing column selection, so I think the user should be aware that their data requires that approach and adjusts on their side intentionally.

On the checking, I have set it up to read the parquet schema when it's possible that we may be doing a partial load ("." in the name) and checking to see if that column is a base column, or if it's a nested column. If it's a nested column then I have it simply return the error if it's not a struct. Notably, this is not going the full distance of verifying that something is a struct-list, which I'm not sure if that's better here or not. But it does handle the main case of catching list-structs.

### DIFF
--- a/tests/nested_pandas/nestedframe/test_io.py
+++ b/tests/nested_pandas/nestedframe/test_io.py
@@ -471,5 +471,5 @@ def test_list_struct_partial_loading_error():
 def test_normal_loading_error():
     """Test that making a normal naming mistake raises the normal pyarrow error."""
     # Load in the example file
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="No match for*"):
         read_parquet("tests/test_data/nested.parquet", columns=["not_a_column"])
--- a/src/nested_pandas/nestedframe/io.py
+++ b/src/nested_pandas/nestedframe/io.py
@@ -229,28 +229,36 @@ def _read_parquet_into_table(
             fs=filesystem,
             engine="pyarrow",
         ) as parquet_file:
-            try:
-                return pq.read_table(parquet_file, columns=columns, **kwargs)
-            except ArrowInvalid as e:
-                if columns is not None:
-                    check_schema = any("." in col for col in columns)  # Check for potential partial loads
-                    if check_schema:
-                        try:
-                            _validate_structs_from_schema(data, columns=columns, filesystem=filesystem)
-                        except ValueError as validation_error:
-                            raise validation_error from e  # Chain the exceptions for better context
-                raise e
+            return _read_table_with_partial_load_check(parquet_file, columns=columns, **kwargs)
 
     # All other cases, including file-like objects, directories, and
     # even lists of the foregoing.
 
     # If `filesystem` is specified - use it, passing it as part of **kwargs
     if kwargs.get("filesystem") is not None:
-        return pq.read_table(data, columns=columns, **kwargs)
+        return _read_table_with_partial_load_check(data, columns=columns, **kwargs)
 
     # Otherwise convert with a special function
     data, filesystem = _transform_read_parquet_data_arg(data)
-    return pq.read_table(data, filesystem=filesystem, columns=columns, **kwargs)
+    return _read_table_with_partial_load_check(data, columns=columns, filesystem=filesystem, **kwargs)
+
+
+def _read_table_with_partial_load_check(data, columns=None, filesystem=None, **kwargs):
+    """Read a pyarrow table with partial load check for nested structures"""
+    try:
+        return pq.read_table(data, columns=columns, **kwargs)
+    except ArrowInvalid as e:
+        # if it's not related to partial loading of nested structures, re-raise
+        if "No match for" not in str(e):
+            raise e
+        if columns is not None:
+            check_schema = any("." in col for col in columns)  # Check for potential partial loads
+            if check_schema:
+                try:
+                    _validate_structs_from_schema(data, columns=columns, filesystem=filesystem)
+                except ValueError as validation_error:
+                    raise validation_error from e  # Chain the exceptions for better context
+        raise e
 
 
 def _validate_structs_from_schema(data, columns=None, filesystem=None):
@@ -317,7 +325,7 @@ def _read_remote_parquet_directory(
                 fs=filesystem,
                 engine="pyarrow",
             ) as parquet_file:
-                table = pq.read_table(parquet_file, columns=columns, **kwargs)
+                table = _read_table_with_partial_load_check(parquet_file, columns=columns, **kwargs)
         tables.append(table)
     return pa.concat_tables(tables)
 


hombit (commented):
Thanks! I'm a little worried about doing one more read just for that. Maybe we can wrap pyarrow's error, so we read the schema only if pyarrow fails with an error about a missing column?

hombit:
@dougbrn yes, I think it would be the perfect solution. I think we can just try to rely on error messages for that. I think it should be fine until we test both lowest and highest pyarrow versions on CI.

dougbrn:
@hombit Now doing the schema check only after a failed read, the result here is that both the original error message and the nested-pandas error message are present

tests/nested_pandas/nestedframe/test_io.py
```diff
@@ -459,3 +459,17 @@ def test__get_storage_options():
     assert storage_opts is not None
     # S3 should NOT have the block_size override (only HTTP/HTTPS)
     assert storage_opts.get("block_size") != FSSPEC_BLOCK_SIZE
+
+
+def test_list_struct_partial_loading_error():
+    """Test that attempting to partially load a list-struct raises an error."""
+    # Load in the example file
+    with pytest.raises(ValueError):
```
hombit: Let's check that it is "our" ValueError by checking the message (`pytest.raises(ValueError, match=...)`). Because `pa.ArrowInvalid` is also `ValueError`

PR #430, opened 2026-01-07
Add is_dir shortcut for skipping directory checks

### LINKED ISSUE #408: Allow skipping directory checks in `read_parquet`
**Feature request**

Since we switched to `fsspec.parquet.open_parquet_file` for our parquet reading code, we added extra round-trips for remote filesystems to check if the destination is a directory. We should allow users to specify if the destination is a directory or a file, so we can skip this check.

I propose to add `read_parquet(..., is_dir: bool | None = None)`, which will perform that check, the same way it is implemented in HATS:
https://github.com/astronomy-commons/hats/pull/601

This would help to resolve this HATS issue:
https://github.com/astronomy-commons/hats/issues/578

**Before submitting**
Please check the following:

- [x] If I have ideas for how the new feature could be implemented, I have provided explanations and/or pseudocode and/or task lists for the steps.

dougbrn:
Closes #408 

Adds an is_dir override that plugs into the `_is_local_dir` and `_is_remote_dir` helper functions, which allows `npd.read_parquet(..., is_dir=True/False)` to apply the override. 

This override isn't absolute for the remote check, as it currently prohibits remote HTTP file directories. I am not sure if providing this override should actually allow HTTP directories with the main concerns being the cost of .is_dir() and potentially iter_dir()

### DIFF
--- a/src/nested_pandas/nestedframe/io.py
+++ b/src/nested_pandas/nestedframe/io.py
@@ -78,9 +78,9 @@ def read_parquet(
         If True, the pointer represents a pixel directory; if False, the pointer
         represents a file. In both cases there is no need to check the pointer's
         content type. If `is_dir` is None (default), this method will resort to
-        `upath.is_dir()` to identify the type of pointer. Inferring the type for
-        HTTP is particularly expensive because it requires downloading the contents
-        of the pointer in its entirety.
+        `upath.is_dir()` to identify the type of pointer. This argument is ignored
+        for HTTP, as inferring the type for HTTP is particularly expensive because
+        it requires downloading the contents of the pointer in its entirety.
     kwargs: dict
         Keyword arguments passed to `pyarrow.parquet.read_table`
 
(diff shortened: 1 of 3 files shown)


src/nested_pandas/nestedframe/io.py
```diff
@@ -73,6 +74,13 @@ def read_parquet(
         Columns specified here will be read using the corresponding pandas.ArrowDtype.
     autocast_list: bool, default=True
         If True, automatically cast list columns to nested columns with NestedDType.
+    is_dir: bool, None, default=None
+        If True, the pointer represents a pixel directory, otherwise, the pointer
```
gitosaurus:
```suggestion
        If True, the pointer represents a pixel directory; if False, the pointer
```
Since `None` is one of the "otherwise" cases.

src/nested_pandas/nestedframe/io.py
```diff
@@ -287,24 +296,36 @@ def _validate_structs_from_schema(data, columns=None, filesystem=None):
...
+
+    # For non-local filesystems, return False right away
+    if upath.protocol not in ("", "file"):
+        return False
+    # For local filesystems, check is_dir if provided, otherwise use upath.is_dir()
+    else:
+        if is_dir is None:
+            return upath.is_dir()
+        else:
+            return is_dir
```
gitosaurus:
```suggestion
        return upath.is_dir() if is_dir is None else is_dir
```

gitosaurus (approved):
LGTM

hombit (approved):
Looks good, thank you. With LSDB, we never really test HTTP directories, and I agree with your approach to that for nested-pandas.

PR #431, opened 2026-01-08
Auto-formatting and clarify testing expectations

delucchi-cmu:
This is a pre-factor PR that makes no meaningful logic changes. I was trying to get another PR ready, and local version skew was changing formatting. 

Also sneaks in a test change to inject the expected value in parametrization.


dougbrn (approved):
LGTM!

PR #433, opened 2026-01-09
Append sub-columns via `map_rows`

### LINKED ISSUE #347: Support adding a new sub-column via `NestedFrame.map_rows()`
**Feature request**

It would be great, if `.reduce()` could add a new sub-column directly to existing nested column. Example with a toy UDF:
```python
from nested_pandas.datasets import generate_dataset

nf = generate_dataset(5, 4)
nf.reduce(lambda flux: {"nested.x": flux - flux.mean()}, "nested.flux", append_columns=True)
```

Currently, this would produce the second "nested" column, while it would be really nice if it would append a new sub-column directly to exiting nested column.

My current workaround is

```python
new = nf.reduce(lambda flux: {"new.x": flux - flux.mean()}, "nested.flux")
nf["nested.x"] = new["new.x"]
```

**Before submitting**
Please check the following:

camposandro:
When using `map_rows(..., append_columns=True)`, allow appending a nested sub-column to an existing `NestedFrame`. Previously, specifying a sub-column for an existing nested column would create a new nested column with the same name.

Closes #347.

### DIFF
--- a/src/nested_pandas/nestedframe/core.py
+++ b/src/nested_pandas/nestedframe/core.py
@@ -2041,7 +2041,11 @@ class NestedFrame(pd.DataFrame):
             will be returned as base columns. Note that this will trigger off of names specified in
             `output_names` in addition to names returned by the user function.
         append_columns : bool, default False
-            if True, the output columns should be appended to those in the original NestedFrame.
+            If True, the output columns are appended to those in the original NestedFrame.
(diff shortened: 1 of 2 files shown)


dougbrn (commented):
Looks good, thank you! Just want to see it added to the docs and then this should be good to go

dougbrn (approved):
Looks great, thank you!

PR #438, opened 2026-01-15
Allow str inputs for series/accessor methods: `explode`, `to_flat`, `to_lists`

### LINKED ISSUE #381: Allow explicit support for `str` inputs to `NestedSeries.explode`, `NestSeriesAccessor.to_flat`, `NestSeriesAccessor.to_lists``
It's natural for all of these to invoke something like:
```
ns.explode("sub_column")
```

but we don't have the explicit string handler (to just store it in a list up front) so strange behavior can happen when trying to do this. Let's add the string handler. Extra motivated by the deprecation of `get_flat_series`, as this becomes more important as a performant replacement

dougbrn:
Fixes #381


camposandro (approved):
Looks good to me!

Issue #428, opened 2026-01-06
Cannot select sub-columns from a particular file


hombit:
**Bug report**

[This file](https://huggingface.co/datasets/MultimodalUniverse/desi/blob/main/edr_sv3/train-00000-of-00007.parquet), both remotely and locally

```
import nested_pandas as npd

npd.read_parquet('train-00000-of-00007.parquet', columns=['spectrum.flux'])
```

<details><summary>traceback</summary>

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[8], line 1
----> 1 npd.read_parquet('train-00000-of-00007.parquet', columns=['spectrum.flux'])

File ~/.virtualenvs/default/lib/python3.12/site-packages/nested_pandas/nestedframe/io.py:133, in read_parquet(data, columns, reject_nesting, autocast_list, **kwargs)
    130 elif isinstance(reject_nesting, str):
    131     reject_nesting = [reject_nesting]
--> 133 table = _read_parquet_into_table(data, columns, **kwargs)
    135 # Resolve partial loading of nested structures
    136 # Using pyarrow to avoid naming conflicts from partial loading ("flux" vs "lc.flux")
    137 # Use input column names and the table column names to determine if a column
    138 # was from a nested column.
    139 if columns is not None:

File ~/.virtualenvs/default/lib/python3.12/site-packages/nested_pandas/nestedframe/io.py:232, in _read_parquet_into_table(data, columns, **kwargs)
    221         return _read_remote_parquet_directory(
    222             path_to_data, filesystem, storage_options, columns, **kwargs
    223         )
    225     with fsspec.parquet.open_parquet_file(
    226         path_to_data.path,
    227         columns=columns,
   (...)    230         engine="pyarrow",
    231     ) as parquet_file:
--> 232         return _read_table_with_partial_load_check(parquet_file, columns=columns, **kwargs)
    234 # All other cases, including file-like objects, directories, and
    235 # even lists of the foregoing.
    236
    237 # If `filesystem` is specified - use it, passing it as part of **kwargs
    238 if kwargs.get("filesystem") is not None:

File ~/.virtualenvs/default/lib/python3.12/site-packages/nested_pandas/nestedframe/io.py:249, in _read_table_with_partial_load_check(data, columns, filesystem, **kwargs)
    247 """Read a pyarrow table with partial load check for nested structures"""
    248 try:
[...]

PR #439, opened 2026-01-16
pin fsspec<2025.12.0 and add issue 428 test

dougbrn:
Allows failing case in #428 to work by pinning fsspec to avoid a recent regression, and adds a direct test that should fail in response to this regression. This shouldn't neccesarily close the issue as we would like to eventually have latest fsspec working, tracker issue for that: https://github.com/fsspec/filesystem_spec/issues/1973


hombit (approved):
Thanks

PR #450, opened 2026-02-05
Improve error message for pack_flat with NaN values in index

### LINKED ISSUE #440: Improve error message for `pack_flat` with nan
**Bug report**

When there are nan in the column to join, `pack_flat` fails:

```
File ~/.local/lib/python3.13/site-packages/nested_pandas/series/packer.py:101, in pack_flat()
     99 # pandas knows when index is pre-sorted, so it would do nothing if it is already sorted
    100 sorted_flat = df.sort_index(kind="stable")
--> 101 return pack_sorted_df_into_struct(sorted_flat, name=name)

File ~/.local/lib/python3.13/site-packages/nested_pandas/series/packer.py:161, in pack_sorted_df_into_struct()
    142 """Make a structure of lists representation of a "flat" dataframe.
    143 
    144 Input dataframe must be sorted and all the columns must have pyarrow dtypes.
   (...)    158     Output series, with unique indexes.
    159 """
    160 if not df.index.is_monotonic_increasing:
--> 161   raise ValueError("The index of the input dataframe must be sorted")
    163 packed_df = view_sorted_df_as_list_arrays(df)
    164 # No need to validate the dataframe, the length of the nested arrays is forced to be the same by
    165 # the view_sorted_df_as_list_arrays function.

ValueError: The index of the input dataframe must be sorted
```

While this seems to be the expected behavior the error message could be clearer. We should check for NaNs and raise an error if they exist.

```
import pandas as pd
>>> pd.Index([1,2,2,3,4], dtype=float).is_monotonic_increasing
True
>>> pd.Index([1,2,2,3,None], dtype=float).is_monotonic_increasing
False
```

**Environment Information**
- nested-pandas 0.6.6
- python 3.13

**Before submitting**
Please check the following:

- [X] I have described the situation in which the bug arose, including what code was executed, and any applicable data others will need to reproduce the problem.
- [X] I have included information about my environment, including the version of this package (e.g. `nested_pandas.__version__`)

hombit:
When the index contains NaN values, is_monotonic_increasing returns
False, causing a misleading "index must be sorted" error. Now we check
for NaN values when monotonicity fails and raise a specific error
explaining that NaN values are not supported for grouping.

Fixes #440

https://claude.ai/code/session_01F7cbsuGUbfDoW1oDdiGqf5

### DIFF
--- a/src/nested_pandas/series/packer.py
+++ b/src/nested_pandas/series/packer.py
@@ -95,29 +95,21 @@ def pack_flat(
     """
 
     if on is not None:
-        if isinstance(on, str):
-            cols = [on]
-        else:
-            cols = list(on)
-        if df[cols].isna().any().any():
-            raise ValueError(
-                f"Column(s) {cols} contain NaN values. "
-                "NaN values are not supported because they cannot be used for grouping rows. "
-                "Please remove or fill NaN values before packing."
-            )
+        cols = [on] if isinstance(on, str) else list(on)
+        for col in cols:
+            if df[col].hasnans:
+                raise ValueError(
+                    f"Column '{col}' contains NaN values. "
+                    "NaN values are not supported because they cannot be used for grouping rows. "
+                    "Please remove or fill NaN values before packing."
+                )
         df = df.set_index(on)
-    else:
-        try:
-            has_nans = df.index.isna().any()
-        except NotImplementedError:
-            # MultiIndex does not support isna()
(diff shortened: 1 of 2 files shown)


dougbrn (approved):
Looks good! Just one initial thought but I think its fine.

dougbrn (commented):
For some reason I can't approve this without leaving a comment, but looks good!

PR #462, opened 2026-03-03
Change internal array representation to `LargeListArray`

### LINKED ISSUE #95: Handle series with more than 2^31 "flat" elements
Currently we do not really care about large datasets, and there are few places where index/offsets computations would fail because of int32 overflow. The reason is that pyarrow uses int32 arrays for offsets. 

There are two possible solutions for this:
- Use `pyarrow.LargeList`
- Smarter and wider usage of chunks.

hombit:
Changes internal `NestedExtensionArray` to use `pa.LargeListArray` (int64 offset) instead of `pa.ListArray` (int32 offset). This is motivated by wanting to support dataframes with more than 2**31 **nested** elements, which may be the case when loading large datasets with `nested-pandas` or returning large results from LSDB with `.compute()`. (I faced it myself when operating with DP2 pilots.)

~This PR introduces breaking changes: by default all outputs are now large lists, including `ndf.nested.to_lists()`, `ndf.to_parquet()`, `pa.array(ndf.nested)`, etc. However, this PR provides a new `large_list: bool = True` argument which, when set to `False`, returns "normal" lists. I'd like to hear opinions on whether we should keep this behavior or set it to `False` by default, from the perspective of `hats`/`hats-import`/`lsdb` usage.~
Changed to `large_list: bool = False` by default, the only case where we have LargeList is `pa.array(ndf.nested)`, but I think it is ok.

The alternative design would be a better support of chunked arrays, because we quite aggressively re-chunk the data in some operations. This would be much harder to implement and test, and also could lead to "memory fragmentation" issues in some use cases (for example, concatenation of dozens of thousands of partitions happening when running `lsdb.Catalog.compute()` over a large catalog).

Closes #95


dougbrn (commented):
I think this looks like a reasonable implementation, I have a couple of thoughts/comments:
* Expectedly, there's a performance hit to this change (~10% on two of our benchmarks), and it sounds like you have use cases where you've run into this, but it does hurt to take a hit like this for all cases because of incompatibility at the edges.
* Have you considered some kind of pandas.options parallel here for us to swap the backend? Probably this is a can of worms, but didn't know if you had thought about it at all.
* As to the default value for `large_list` kwarg, I don't know I could see arguments for both. I liked `False` initially for minimal disruption of potential downstream workflows, but not sure if invoking the downcast hits performance at all in these cases? Default `True` seems nice in that the only reason someone who try to move off of it would be for fine-tuning performance (again if that even provides a benefit), or downstream compatibility.

hombit:
After the discussions we go with this approach, but with `large_list=False` by default (with better errors when it would fail)

src/nested_pandas/series/utils.py
```diff
@@ -63,7 +63,102 @@ def is_pa_type_is_list_struct(pa_type: pa.DataType) -> bool:
...
+            )
+
+        list_arrays.append(zero_align_large_list_offsets(list_array))
+    return pa.StructArray.from_arrays(
+        arrays=list_arrays,
+        fields=struct_fields(array.type),
+    )
+
+
+def zero_align_offsets(array: pa.LargeListArray | pa.StructArray) -> pa.LargeListArray | pa.StructArray:
```
dougbrn: I noticed that this doesn't have test coverage, would be nice to add to not degrade project coverage too much

dougbrn (approved):
Overall looks great, thanks for doing so much digging on this! Just one nit to up code coverage which is not a blocking request if you don't have much time

src/nested_pandas/series/utils.py
hombit: Nice catch, will do!

hombit: Actually that function was added by a mistake, it is part of an implementation which was never finished. So I've just removed it!

PR #474, opened 2026-03-22
feat: export from_pyarrow as top-level function and add docstring example

### LINKED ISSUE #432: Properly seat `from_pyarrow` as an I/O function
Add it to our docs, and make it importable directly through the class (e.g. npd.from_pyarrow)

tanisha-raha:
Fixes #432

## Changes
- Added `from_pyarrow` to the top-level `__init__.py` imports so it can be used as `npd.from_pyarrow()`
- Added `from_pyarrow` to `__all__`
- Added a docstring example to `from_pyarrow` in `io.py`

## Testing
All 412 existing tests pass with no failures.

### DIFF
--- a/src/nested_pandas/__init__.py
+++ b/src/nested_pandas/__init__.py
@@ -1,10 +1,10 @@
 from ._version import __version__
 from .nestedframe import NestedFrame
-from .nestedframe.io import read_parquet
+from .nestedframe.io import read_parquet, from_pyarrow
 
 # Import for registering
 from .series.accessor import NestSeriesAccessor  # noqa: F401
 from .series.dtype import NestedDtype
 from .series.nestedseries import NestedSeries
 
-__all__ = ["NestedDtype", "NestedFrame", "read_parquet", "NestedSeries", "__version__"]
+__all__ = ["NestedDtype", "NestedFrame", "read_parquet", "from_pyarrow", "NestedSeries", "__version__"]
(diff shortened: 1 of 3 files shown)


src/nested_pandas/nestedframe/io.py
```diff
@@ -448,6 +448,13 @@ def from_pyarrow(
     -------
     NestedFrame
 
+    Examples
+    --------
+    >>> import nested_pandas as npd
+    >>> import pyarrow as pa
+    >>> table = pa.table({"a": [1, 2, 3], "b": ["x", "y", "z"]})
```
delucchi-cmu: Use test data that contains some nested fields, to better showcase the feature.

src/nested_pandas/__init__.py
```diff
@@ -1,10 +1,10 @@
 from ._version import __version__
 from .nestedframe import NestedFrame
-from .nestedframe.io import read_parquet
+from .nestedframe.io import read_parquet, from_pyarrow
```
delucchi-cmu: This should also be exposed on the `API Reference` I/O section of the [nested-pandas documentation](https://nested-pandas.readthedocs.io/en/latest/reference/nestedframe.html).

`docs/reference/nestedframe.rst`

delucchi-cmu: Fix the import order with `ruff` - this causes a pre-commit failure.

delucchi-cmu (approved):

PR #475, opened 2026-03-24
Chunking and rechunking functionality for large datasets

### LINKED ISSUE #95: Handle series with more than 2^31 "flat" elements

hombit:
This is an alternative approach to support large nested arrays, >2**31 nested values; see #462 for another approach. It is aimed at two things: 1) prevent `offsets` overflow, and 2) prevent small chunks appearance and memory borrowing with re-chunking. In contrast to #462, this is not a breaking change, but it applies a lot of trade-offs and tuning hyperparameters, which are not validated for the best performance.

Closes #95

### DIFF
--- a/src/nested_pandas/series/accessor.py
+++ b/src/nested_pandas/series/accessor.py
@@ -171,7 +171,7 @@ class NestSeriesAccessor(Mapping):
     @property
     def flat_index(self) -> pd.Index:
         """Index of the flattened arrays"""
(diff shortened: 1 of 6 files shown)


hombit:
We chose approach implemented in #462

PR #476, opened 2026-03-26
DONT-MERGE GSOC26: Improve performance of replace_with_mask()

### LINKED ISSUE #52: Improve performance of series.ext_array.replace_with_mask()
Currently, arrow misses the support of `pyarrow.compute.replace_with_mask` for struct arrays:
https://github.com/apache/arrow/issues/29558

That's why we have our own implementation used by `NestedExtenstionArray.__setitem__()`. The implementation has an overhead of creating a `len(self)`-sized struct array to perform the replacement. This approach would work well when we are going to replace many elements, but when we replacing just few, it would produce a large memory foot-print and probably take a while.

An alternative approach would be copying the original array to `np.ndarray[pa.StructScalar]`, replace the elements in-place, and convert it back:
```python
def replace_with_mask(array: pa.ChunkedArray, mask: pa.BooleanArray, value: pa.Array) -> pa.ChunkedArray:
    """Replace the elements of the array with the value where the mask is True"""
    np_array = np.fromiter(array, dtype=object)
    np_array[mask] = value
    new_pa_array = pa.array(np_array)
    return pa.chunked_array([new_pa_array])
```

We should create a benchmark and see what works faster and have smaller memory foot-print.

s7g4:
## Change Description
Closes #52

Optimized `NestedExtensionArray.__setitem__` and the internal `replace_with_mask` utility to handle `pyarrow.Scalar` values directly. This avoids the significant overhead caused by expanding a single replacement scalar into a full-sized `pyarrow.Array` (matching the length of the target array).

## Solution Description
The optimization involves two main changes in `src/nested_pandas/series/ext_array.py`:
1.  **`NestedExtensionArray.__setitem__`**: Now detects if the replacement value can be boxed as a `pa.Scalar`. If so, it passes the scalar directly instead of creating a `len(self)`-sized array.
2.  **`replace_with_mask`**: Updated to handle `pa.Scalar` inputs. When a scalar is detected, it leverages PyArrow's native `pa.compute.if_else` kernel for efficient broadcasting.

**Benchmarks (1,000,000 rows):**
- **Original approach:** ~7.3s execution time | ~78MB peak memory
- **Optimized approach:** **~0.26s** execution time | **~0MB** peak memory (approx. **30x faster**)

## Code Quality
- [x] I have read the Contribution Guide and agree to the Code of Conduct

### DIFF
--- a/src/nested_pandas/series/ext_array.py
+++ b/src/nested_pandas/series/ext_array.py
@@ -177,10 +177,11 @@ def to_pyarrow_dtype(
     return None
 
 
-def replace_with_mask(array: pa.ChunkedArray, mask: pa.BooleanArray, value: pa.Array) -> pa.ChunkedArray:
+def replace_with_mask(array: pa.ChunkedArray, mask: pa.BooleanArray, value: pa.Array | pa.Scalar) -> pa.ChunkedArray:
     """Replace the elements of the array with the value where the mask is True"""
-    # TODO: performance optimization
-    # https://github.com/lincc-frameworks/nested-pandas/issues/52
+    # Fast path for scalars using native broadcasting
+    if isinstance(value, pa.Scalar):
+        return pa.compute.if_else(mask, value, array)
 
     # If mask is [False, True, False, True], mask_cumsum will be [0, 1, 1, 2]
     # So we put value items to the right positions in broadcast_value, while duplicate some other items for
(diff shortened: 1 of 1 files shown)


hombit:
Performance gains look great, but changes are not tested. Closing for now, feel free to re-submit with more tests.

PR #480, opened 2026-03-31
Fix #460: read_parquet(use_pandas_metadata: bool = True)

### LINKED ISSUE #460: Allow ignore_metadata when creating pandas
**Bug report**

I've been debugging the following LSDB issue: https://github.com/astronomy-commons/lsdb/issues/1265

Here's what I think is happening. 

The data is older (imported with hats v0.4.0), and still contains pandas metadata inside the parquet files. While newer files will have this metadata suppressed, it's not a guarantee that these will be missing.

This automatically sets the index to `_healpix_29`, based on the pandas metadata. This is also true in the `_ddf.meta`.

I would like `npd.read_parquet` to return to me a nested frame that has no index set, and I believe that's not currently possible.

nested_pandas version: '0.6.8'

**Before submitting**
Please check the following:

- [x] I have described the situation in which the bug arose, including what code was executed, and any applicable data others will need to reproduce the problem.
- [x] I have included information about my environment, including the version of this package (e.g. `nested_pandas.__version__`)
- [x] I have included available evidence of the unexpected behavior (including error messages, screenshots, and/or plots) as well as a description of what I expected instead.

hombit:
Adds `read_parquet(use_pandas_metadata: bool = True)`, which allows to ignore index and other pandas metadata.

From LSDB/hats side we should pass `use_pandas_metadata=False` to address #460 and https://github.com/astronomy-commons/lsdb/issues/1265

Closes #460

### DIFF
--- a/src/nested_pandas/nestedframe/io.py
+++ b/src/nested_pandas/nestedframe/io.py
@@ -85,8 +85,8 @@ def read_parquet(
     ignore_pandas_metadata: bool, default=True
         If True (default), ignore the pandas metadata stored in the Parquet file's
         schema when constructing the NestedFrame. This prevents the index from being
-        set based on the pandas metadata, and avoids other inconsistencies introduced
-        by stale or unexpected metadata embedded in older Parquet files.
+        set based on the pandas metadata, and avoids column dtype casting driven by
+        stale pandas metadata embedded in older Parquet files.
 
         .. note::
             This differs from the default pandas behavior (pd.read_parquet),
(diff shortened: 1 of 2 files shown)


dougbrn (commented):
The addition of this flag seems reasonable. I'm definitely wondering about having the default be True, from the perspective of LSDB it makes sense but from the perspective of nested-pandas being a pandas extension it seems weird to turn off our reading of pandas metadata by default.

dougbrn (approved):
Thank you!

PR #486, opened 2026-04-09
Nicer HTML repr for small nested elements

hombit:
Change representation for small nested elements: one or two nested rows

**Before:**
Screenshot 2026-04-09 at 14 20 35

**After:**
Screenshot 2026-04-09 at 14 20 57


olivialynn (approved):
Looks great!

PR #503, opened 2026-07-07
Pandas v3 Compatibility

### LINKED ISSUE #441: Bump pandas to v3.0
`pandas` v3.0.0 was released. We should upgrade
https://pandas.pydata.org/docs/whatsnew/v3.0.0.html
https://github.com/pandas-dev/pandas/releases/tag/v3.0.0
https://github.com/pandas-dev/pandas/actions/runs/21212167278

### LINKED ISSUE #444: Drop Python 3.10 Support; Add 3.14
We recently had discussion about aligning LF projects under a similar rolling window of python version compatibility. With LSDB no longer supporting 3.10, that would suggest we should roll forward as well

dougbrn:
Closes #441 

Pandas 3.0 requires Python 3.11+, so also closes #444 

Changes:
generate_dataset’s band column is now a large_string rather than a string, I changed the doctests where it was failing, but perhaps that should be fixed on generate_datasets side

pandas 3.0 changed `ArrowExtensionArray._box_pa_array` (used internally by `Series.astype`/`pd.Series(..., dtype=...)`). Previously it called `pyarrow.array(value, type=pa_type)`, which forwards `type` into your `NestedExtensionArray.__arrow_array__(type=...)` — letting it pick struct-array vs list-array representation correctly. In pandas 3.0, a new branch calls `value.__arrow_array__()` with **no type**, then does a bare `pyarrow.Array.cast(pa_type)` afterward. Since your struct-of-lists and list-of-structs representations are transposes of each other, pyarrow's native cast kernel can't convert between them — hence `ArrowNotImplementedError: Unsupported cast from large_list<...> to struct`.

Future Feature Requests:

pandas 3 adds pd.col, which we should investigate for potential nested column compatibility/API behavior. Issue here #504

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.11', '3.12', '3.13', '3.14']
 
     steps:
     - uses: actions/checkout@v7
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.11', '3.12', '3.13']
+        python-version: ['3.11', '3.12', '3.13', '3.14']
 
     steps:
     - uses: actions/checkout@v7
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.11', '3.12', '3.13', '3.14']
+        python-version: ['3.11', '3.12', '3.13']
 
     steps:
     - uses: actions/checkout@v7
--- a/.github/workflows/testing-and-coverage.yml
+++ b/.github/workflows/testing-and-coverage.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.11', '3.12', '3.13', '3.14']
 
     steps:
     - uses: actions/checkout@v7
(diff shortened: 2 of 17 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.11', '3.12', '3.13']
```
hombit:
```suggestion
        python-version: ['3.11', '3.12', '3.13', '3.14']
```

.github/workflows/testing-and-coverage.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.11', '3.12', '3.13']
```
hombit:
```suggestion
        python-version: ['3.11', '3.12', '3.13', '3.14']
```

src/nested_pandas/series/ext_array.py
hombit, line 1: According to the [pandas v3 changelog](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v3.0.0.html) we should make this change:
- Change `_from_sequence_of_strings(..., dtype: ExtensionDtype, ...)` https://github.com/pandas-dev/pandas/pull/56519/changes#diff-4ea2067eb8de5ace5b5a9e8b6673492f801b819a72b2be9380e35cdd7eb194c9R321

pyproject.toml
```diff
@@ -16,14 +16,14 @@ classifiers = [
     "Programming Language :: Python",
 ]
 dynamic = ["version"]
-requires-python = ">=3.10"
+requires-python = ">=3.11"
 dependencies = [
     "numpy>=2",
     # We use internal pd._libs.missing and experimental ArrowExtensionArray
-    "pandas>=2.2.3,<2.4",
+    "pandas>=3.0.0",
```
hombit:
```suggestion
    "pandas>=3,<4",
```

hombit (approved):
Great work, thank you! I request a single small change and suggest a few others.

src/nested_pandas/series/ext_array.py
dougbrn: Done!

Issue #521, opened 2026-08-27
Using ffspec is slow when used on local files


nevencaplar:
Reading local files with ffspec.parquet is roughly 2x slower than reading with plain pyarrow. I believe that changes in #385 put ffspec to be used for both remote and local files; but effectively this introduces large overhead. It seems that this article was the motivation: https://developer.nvidia.com/blog/optimizing-access-to-parquet-data-with-fsspec/, but this only applies to remote systems. Also it turns out we dont have any benchmarks which would test larger reading of files in e.g., lsdb or hats testing suite. I believe this is part of explanation (biggest part) for https://github.com/astronomy-commons/lsdb/issues/1010. 

I have small notebook to show the difference [at the gist here](https://gist.github.com/nevencaplar/0ccef4dc914f109f082c475a7be9755a). It is vibe coaded, but I tested it in multiple ways and I believe the main message is correct. 

    Loading, 4 order-5 Gaia partitions, 4 of 152 columns, best of 9 :                                                                                      
  
                                                       1 thread  8 threads   bytes read  vs pyarrow
    pq.read_table                    (plain pyarrow)     0.09 s     0.03 s       126 MB       1.00x
    npd.read_parquet   (what hats calls)                 0.17 s     0.07 s       132 MB       1.94x

hombit:
While I agree with the raised concern, I think that the benchmark methodology is wrong and you are actually reading from the RAM disk, not from the actual FS. This is why disk read benchmarks are hard. However, we can implement this kind of RAM-disk benchmarks, and they should be pretty stable in the CI.

nevencaplar:
Right, because Claude is really wanting to do - what it calls ``warm`` run - and ``warm`` meaning loading from RAM...

PR #522, opened 2026-08-28
Parquet read benchmark for tmpfs

hombit:
It adds a RAM-based benchmark for local data reads as proposed by #521. Locally, I can reproduce the difference between `nested-pandas` and `pyarrow` with this benchmark, but it is not as dramatic as in #521, something like 50% slower.

### DIFF
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -3,6 +3,9 @@
 For more information on writing benchmarks:
 https://asv.readthedocs.io/en/stable/writing_benchmarks.html."""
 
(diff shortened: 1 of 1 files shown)


smcguire-cmu (approved):
This looks good, but I think we should also have one that tests cold storage performance as well as page cached

hombit:
@smcguire-cmu I'm not sure if CI is stable enough for that. I've opened an issue to experiment with cold reads later https://github.com/lincc-frameworks/nested-pandas/issues/527

PR #523, opened 2026-08-28
Optimize local parquet reads

### LINKED ISSUE #521: Using ffspec is slow when used on local files

hombit:
From https://github.com/lincc-frameworks/nested-pandas/actions/runs/33193355245/job/98924190522#step:10:1
```
| Change   | Before [b85781ba]    | After [86d61ad0]    |   Ratio | Benchmark (Parameter)                   |
|----------|----------------------|---------------------|---------|-----------------------------------------|
| -        | 58.0±2ms             | 39.5±2ms            |    0.68 | benchmarks.ReadFewColumnsTmpfs.time_run |
```

Fixes #521

### DIFF
--- a/src/nested_pandas/nestedframe/io.py
+++ b/src/nested_pandas/nestedframe/io.py
@@ -220,21 +220,21 @@ def _read_parquet_into_table(
 ) -> pa.Table:
     """Reads parquet file(s) from path and returns a pyarrow table.
 
-    For single Parquet file paths, we want to use
-    `fsspec.parquet.open_parquet_file`.  But for any other usage
-    (which includes file-like objects, local directories and lists
-    thereof), we want to defer to `pq.read_table`.
+    For single remote Parquet file paths, we want to use
+    `fsspec.parquet.open_parquet_file`. Remote directories are handled
+    separately via `_read_remote_parquet_directory`. For everything local
+    (files and directories alike), as well as file-like objects and lists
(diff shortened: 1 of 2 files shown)


dougbrn (approved):
LGTM!

PR #526, opened 2026-08-28
Use fsspec.parquet.open_parquet_files for remote directories

hombit:
`fsspec.parquet.open_parquet_files` provides a more optimal interface for reading multiple paths from the same remote filesystem


dougbrn (approved):
Nice, seems like a good change

PR #531, opened 2026-09-01
Use `datafusion` as an optional `read_parquet` engine

hombit:
Adds initial and optional support for `datafusion` as a Parquet reader engine. In certain cases, `datafusion` may be 10–50 times faster than `pyarrow`, for instance, when selecting a few rows only from a file with page statistics or when selecting up to ~20% of nested subcolumns.

Further improvements will be required to automatically switch engine for better performance.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,9 @@ classifiers = [
 dynamic = ["version"]
 requires-python = ">=3.11"
 dependencies = [
+    "datafusion>=43",  # adds parquet.schema_force_view_types
+    # datafusion requires it, and versions < 4 conflict with other stuff
+    "typing-extensions>=4.0.0; python_version < '3.13'",
     "numpy>=2",
     # We use internal pd._libs.missing and experimental ArrowExtensionArray
     "pandas>=3,<4",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ requires-python = ">=3.11"
 dependencies = [
     "datafusion>=43",  # adds parquet.schema_force_view_types
     # datafusion requires it, and versions < 4 conflict with other stuff
-    "typing-extensions>=4.0.0; python_version < '3.13'",
+    "typing-extensions>=4.5.0; python_version < '3.13'",
     "numpy>=2",
     # We use internal pd._libs.missing and experimental ArrowExtensionArray
     "pandas>=3,<4",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,8 +18,8 @@ classifiers = [
 dynamic = ["version"]
 requires-python = ">=3.11"
 dependencies = [
-    "datafusion>=43",  # adds parquet.schema_force_view_types
-    # datafusion requires it, and versions < 4 conflict with other stuff
+    # datafusion is an optional dependency of `read_parquet(engine="datafusion")`
+    # typing-extensions versions < 4 conflict with other stuff
     "typing-extensions>=4.5.0; python_version < '3.13'",
     "numpy>=2",
     # We use internal pd._libs.missing and experimental ArrowExtensionArray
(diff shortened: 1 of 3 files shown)


src/nested_pandas/nestedframe/io.py
```diff
@@ -235,6 +265,10 @@ def _read_parquet_into_table(
     lot of "junk" besides of the actual parquet files.
     """
     if isinstance(data, str | Path | UPath) and not _is_local_path(path_to_data := UPath(data)):
+        if engine == "datafusion":
+            # Raises ValueError
```
dougbrn: debug comment?

src/nested_pandas/nestedframe/io.py
```diff
@@ -252,22 +286,130 @@ def _read_parquet_into_table(
...
+    # Both [(column, op, value), ...] and [[(column, op, value), ...], ...]
+    for filter_ in filters:
+        conjunct = [filter_] if isinstance(filter_, tuple) else filter_
+        if not isinstance(conjunct, list | tuple):
+            continue
+        for term in conjunct:
+            if not isinstance(term, tuple) or len(term) != 3:
+                continue
+            column = term[0]
+            if isinstance(column, str) and "." in column:
```
dougbrn, line 313: You could consider checking more directly that this is a known subcolumn: https://nested-pandas.readthedocs.io/en/latest/reference/api/nested_pandas.NestedFrame.get_subcolumns.html#nested_pandas.NestedFrame.get_subcolumns

We don't specifically support dot names in non nested columns, so just a suggestion, but this might be a more accurate approach

src/nested_pandas/nestedframe/io.py
```diff
@@ -284,6 +426,207 @@ def _read_table_with_partial_load_check(data, columns=None, filesystem=None, **k
...
+    table = df.to_arrow_table()
+
+    if columns is not None:
+        table = table.rename_columns([column.split(".")[-1] for column in columns])
+
+    return table
+
+
+# DataFusion session settings for every read.
+DATAFUSION_SESSION_SETTINGS = {
```
dougbrn, line 518: tempted to suggest this be kicked out into a new file/module, potentially along with more of the datafusion config/admin. But don't feel too strongly about it.

pyproject.toml
```diff
@@ -18,6 +18,9 @@ classifiers = [
 dynamic = ["version"]
 requires-python = ">=3.11"
 dependencies = [
+    "datafusion>=43",  # adds parquet.schema_force_view_types
```
dougbrn: should we make this optional?

dougbrn (commented):
Generally, I think this looks good. Have a couple of low-priority things that are non-blocking. But my main question is on whether we should make this an optional dependency for now, it might be nice to not add another dependency to the lsst stack, but additionally it might be simpler to just have it be required (or it might already be a dependency of lsst?)

src/nested_pandas/nestedframe/io.py
hombit: @dougbrn the problem is that we don't necessarily know columns yet at this point. We can optimize by passing schema (if provided by caller) to this function, but I'm not sure if it is worth it.

pyproject.toml
hombit: Yes, we should! I've forgotten about it

src/nested_pandas/nestedframe/io.py
hombit: Good point, but could we postpone it to a following PR? I wait for the next DataFusion version to improve performance. I've opened #549 for it.

src/nested_pandas/nestedframe/io.py
dougbrn: Ah, yes fair point. This is fine then!

src/nested_pandas/nestedframe/io.py
dougbrn: sounds good.

dougbrn (approved):
Looks good, thank you!

Issue #534, opened 2026-09-04
Pandas Extension dtype and array for Tensors


smcguire-cmu:
Using Arrow FixedShapeTensor canonical extension type as the underlying data structure, make a pandas extensiondtype and extensionarray supporting tensor columns.

smcguire-cmu:
Closed in #553

PR #550, opened 2026-09-09
Add Tensor Extension Dtype

smcguire-cmu:
First part of #534, adding the pandas extension dtype for tensors wrapping the pyarrow `fixed_shape_tensor` type.  The code is heavily inspired by the nested extension dtype.

Doesn't support permutation due to limitations of arrow array slicing and the requirement for rows to be the first dimension meaning that any other layout (e.g. fortran) can only be per row.

Some functions unimplemented since missing extension array implementation, but I wanted to split up the PRs to make It easier to review.


src/nested_pandas/tensors/dtype.py
```diff
@@ -0,0 +1,246 @@
...
+        TypeError
+            If the string is not a valid tensor type string, if the element
+            type cannot be parsed, or if the shape and dim_names are
+            inconsistent.
+        """
+        if not isinstance(string, str):
+            raise TypeError(f"'construct_from_string' expects a string, got {type(string)}")
+        match = _NAME_PATTERN.match(string)
+        if match is None:
+            raise TypeError(f"Cannot construct a '{cls.__name__}' from '{string}'")
```
dougbrn: could add a bit more to this error, explaining the expected string format

dougbrn (approved):
Looks great, really clean. Just one nit on adding more to an error output that I think is non-blocking

src/nested_pandas/tensors/dtype.py
```diff
@@ -0,0 +1,246 @@
...
+            pyarrow_dtype = pyarrow_dtype.pyarrow_dtype
+        if not isinstance(pyarrow_dtype, pa.FixedShapeTensorType):
+            raise TypeError(
+                f"TensorDtype can only be constructed with pa.FixedShapeTensorType, got {pyarrow_dtype!r}"
+            )
+        # A permutation reorders the inner dimensions of each tensor in storage. Supporting it would
+        # require a permutation branch in every array method, and pyarrow would return non-contiguous
+        # transposed views, breaking the "every element has dtype.shape" invariant. Columns with a
+        # permutation can be normalized to C order (one copy) and rebuilt without it.
+        if pyarrow_dtype.permutation is not None:
```
hombit: `permutation == list(range(len(pyarrow_dtype.shape)))` should be equivalent to `None`?

hombit (approved):

src/nested_pandas/tensors/dtype.py
smcguire-cmu: It should be equivalent, but I kinda don't want to get into dealing with if they hash the same or compare equal in pyarrow, if we should be stripping the parameter for this specific case, why a user might not see things being equal, etc. Especially since I'm not sure when that might ever be set or actually apply. But if you think it's worth adding we can.

src/nested_pandas/tensors/dtype.py
smcguire-cmu: Good call, expanded it. The error now shows the expected format and two examples:

```
Cannot construct a 'TensorDtype' from 'tensor[double]'. Expected a string like 'tensor[<element type>, (<shape>)]', optionally followed by ', dim_names=[<names>]', for example 'tensor[double, (2, 3)]' or 'tensor[int32, (4, 4), dim_names=[y, x]]'
```

The leading "Cannot construct a 'TensorDtype' from '...'" part is kept verbatim since pandas' extension test suite matches on it.

src/nested_pandas/tensors/dtype.py
smcguire-cmu: Ok I looked into it a bit more and it turns out pyarrow's `from_numpy_ndarray` always sets an identity permutation. Permutation=None vs identity permutation do equal each other, but do hash differently which is a problem since our dtype relies on that hash as the pandas _metadata. So I changed it to accept the identity permutation but set it to None on init.

PR #551, opened 2026-09-10
Use fixtures for test data

smcguire-cmu:
Replaces paths to test data with pytest fixtures. I think this makes the test code cleaner and without repeated paths, and constructs the paths in conftest based on __file__ meaning that the tests now work correctly when run from any working directory.


dougbrn (commented):
Nice, looks good. The failing windows test looks real potentially, so would be good to resolve that first.

smcguire-cmu:
Ah yeah, I didn't see that! It seems like the pyarrow filesystem returns paths on windows with a forward slash since it normalizes it to the path form it expects, but python Path uses backslash on windows. I think the paths all work with the rest of the code, it's just the test comparing them as strings doesn't take into account the difference.

dougbrn (approved):
Thanks, looks great!

Issue #552, opened 2026-09-11
Support reading FixedShapeTensor pyarrow columns as TensorExtensionArray


smcguire-cmu:
Currently pyarrow FixedShapeTensor columns are mapped to pandas ArrowExtensionArray columns (`pd.ArrowDtype`) by `read_parquet` and `from_pyarrow`, rather than to our TensorExtensionArray.

Suggested change: replace `types_mapper=pd.ArrowDtype` in `from_pyarrow` with a mapper that returns `TensorDtype` for `pa.FixedShapeTensorType` and `pd.ArrowDtype` otherwise. `TensorDtype.__from_arrow__` already handles the rest. Round-trip tests through both the pyarrow and datafusion engines, including a file whose tensor type has an identity permutation (what `pa.FixedShapeTensorArray.from_numpy_ndarray` writes; the dtype normalises it away).

Follows on from #534 and is a prerequisite for #544.

Issue #542, opened 2026-09-04
Support NestedFrame setitem with TensorSeries or Numpy array for tensors


smcguire-cmu:
Allow e.g. `ndf["image"] = some_numpy_array` or `ndf["image"] = some_tensor_series`

Issue #539, opened 2026-09-04
TensorSeries - setitem with numpy array


smcguire-cmu:
Support setting an element of a TensorSeries with a numpy array to update an element.

PR #553, opened 2026-09-11
Add Tensor Extension Array

smcguire-cmu:
Adds the `TensorExtensionArray` class with the array for the TensorDType. Inspired heavily by the nested extension array.

Some points about it:
- Series-level assignment of a single tensor via loc/iloc, broadcasting, and fillna with a tensor don't work, because pandas inspects ndarray values before the extension array is consulted. Plain `series[label] = tensor`, sequences, and everything at the array level work. This is what #539 and #542 will address at the TensorSeries/NestedFrame level.
- unique, value_counts, factorize, merge and hashing don't work, since ndarrays aren't hashable. Same as nested with DataFrames.
- Nullable tensor columns don't round-trip Parquet before pyarrow 26.0.0 (apache/arrow#35692 and apache/arrow#35697, fixed 2026-07-28). Verified working on the 26.0.0 nightly and failing on 25.0.1. The round-trip test drops nulls until then.
- Non-identity permutations are rejected; identity ones are normalised away. pyarrow's own `FixedShapeTensorArray.from_numpy_ndarray` always stamps an identity permutation, so it has to be accepted, and pyarrow hashes those types differently despite comparing them equal, which is why it is normalised rather than just tolerated.
- Element-level nulls inside a tensor (which only arrow input can produce) read as NaN, upcasting integer tensors to float, as numpy and pandas do for missing numeric data. Row-level nulls are `pd.NA` as usual.
- read_parquet integration is to come later, new issue #552 for this.
- Some methods in nested were overwriting superclass methods just to call the method and raise errors for not being implemented. I dropped those since they're inherited anyway: they only called `super()`, which raises `AbstractMethodError`, so behaviour is identical with or without them. If they have a reason to be there I can add them back.
- `replace_with_mask` moved from `series/ext_array.py` to `series/utils.py` unchanged, since both arrays use it.
- `pyproject.toml`: mypy `mypy_path`/`explicit_package_bases` so the two `conftest.py` files under `tests/` don't collide.

Testing  runs pandas' extension-array conformance suite (`pandas.tests.extension.base`, 11 suites) alongside individual tests mirroring nested's. The 54 xfails are listed by reason in `tests/nested_pandas/tensors/conftest.py`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,9 @@ build-backend = "setuptools.build_meta"
 
 [tool.mypy]
 python_version = "3.10"
+# Resolve module names from src/ and tests/ so the two conftest.py files do not collide
+mypy_path = "src"
+explicit_package_bases = true
 
 [tool.setuptools_scm]
 write_to = "src/nested_pandas/_version.py"
--- a/src/nested_pandas/tensors/ext_array.py
+++ b/src/nested_pandas/tensors/ext_array.py
@@ -362,6 +362,50 @@ class TensorExtensionArray(ExtensionArray):
         for scalar in self._pa_array:
             yield self._scalar_to_numpy(scalar)
 
+    def __eq__(self, other):  # type: ignore[override]
+        """Elementwise equality with another array of tensors, or with a single tensor.
+
+        Two tensors are equal when they have the same dtype and all their values
+        are equal, with numpy's semantics for the values (so NaN != NaN). The
+        result is NA where either side is missing. Comparing to anything that
+        cannot be interpreted as tensors of this dtype, including an array of a
+        different tensor dtype, gives False everywhere.
+
+        Returns
+        -------
+        pd.arrays.BooleanArray
+        """
+        # Let pandas unbox these and dispatch back to us with the underlying array
+        if isinstance(other, pd.Series | pd.Index | pd.DataFrame):
+            return NotImplemented
+
+        self_mask = self.isna()
+        other_array = self._coerce_comparison_operand(other)
+        if other_array is None:
+            return pd.arrays.BooleanArray(np.zeros(len(self), dtype=bool), self_mask)
+        if len(other_array) not in (1, len(self)):
+            raise ValueError("Lengths must match to compare")
+
+        # Compare the flat values row by row; a length-1 operand broadcasts
+        left = self._values_block().reshape(len(self), self.dtype.size)
+        right = other_array._values_block().reshape(len(other_array), self.dtype.size)
+        equal = np.all(left == right, axis=1)
+        mask = self_mask | other_array.isna()
+        return pd.arrays.BooleanArray(equal, mask)
+
+    def _coerce_comparison_operand(self, other: Any) -> Self | None:  # type: ignore[name-defined] # noqa: F821
+        """Convert the other side of a comparison to an array of this dtype, or None if impossible."""
+        if _is_na(other):
+            return type(self)._from_sequence([None], dtype=self.dtype)
+        if isinstance(other, TensorExtensionArray):
+            return other if other.dtype == self.dtype else None
+        if isinstance(other, np.ndarray) and other.shape == self.dtype.shape:
+            other = [other]
+        try:
+            return type(self)._from_sequence(other, dtype=self.dtype)
+        except (TypeError, ValueError, pa.ArrowInvalid, pa.ArrowTypeError):
+            return None
+
     def to_numpy(
         self,
         dtype: DTypeLike | None = None,
@@ -718,6 +762,22 @@ class TensorExtensionArray(ExtensionArray):
         np.ndarray
             Array of shape ``(len(self), *self.dtype.shape)``.
         """
+        stack = self._values_block()
+        if not self._hasna:
+            return stack
+
+        # The block holds arbitrary values for null tensors, so fill them here
+        result_dtype = np.result_type(stack.dtype, np.min_scalar_type(na_value))
+        result = np.array(stack, dtype=result_dtype)
+        result[self.isna()] = na_value
+        return result
+
+    def _values_block(self) -> np.ndarray:
+        """Read-only numpy view of shape ``(n, *shape)`` over the arrow buffers.
+
+        Zero-copy for a single chunk. Null tensors are present in the block
+        but hold arbitrary values, so callers must consult :meth:`isna`.
+        """
         if len(self) == 0:
             return np.empty((0, *self.dtype.shape), dtype=_numpy_value_dtype(self.dtype))
 
@@ -726,15 +786,7 @@ class TensorExtensionArray(ExtensionArray):
             combined = cast(pa.FixedShapeTensorArray, self._pa_array.chunk(0))
         else:
             combined = cast(pa.FixedShapeTensorArray, self._pa_array.combine_chunks())
-        stack = combined.to_numpy_ndarray()
-        if combined.null_count == 0:
-            return stack
-
-        # to_numpy_ndarray() silently returns garbage for null tensors, so fill them here
-        result_dtype = np.result_type(stack.dtype, np.min_scalar_type(na_value))
-        result = np.array(stack, dtype=result_dtype)
-        result[self.isna()] = na_value
-        return result
+        return combined.to_numpy_ndarray()
 
     @classmethod
     def from_arrow_ext_array(cls, array: ArrowExtensionArray, *, dtype: TensorDtype | None = None) -> Self:  # type: ignore[name-defined] # noqa: F821
--- a/src/nested_pandas/tensors/ext_array.py
+++ b/src/nested_pandas/tensors/ext_array.py
@@ -1,0 +1,741 @@
+# This code is massively adopted from pandas' ArrowExtensionArray. Pandas license is required for this code:
+#
+# BSD 3-Clause License
+#
+# Copyright (c) 2008-2011, AQR Capital Management, LLC, Lambda Foundry, Inc. and PyData Development Team
+# All rights reserved.
+#
+# Copyright (c) 2011-2024, Open source contributors.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+#
+# * Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+#
+# * Neither the name of the copyright holder nor the names of its
+#   contributors may be used to endorse or promote products derived from
+#   this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""Pandas extension array for fixed-shape tensor columns."""
+
+from __future__ import annotations  # Self in Python 3.10
+
+from collections.abc import Callable, Iterator, Sequence
+from typing import Any, cast
+
+import numpy as np
+import pandas as pd
+import pyarrow as pa
+import pyarrow.compute as pc
+from numpy.typing import DTypeLike
+from pandas.api.extensions import no_default
+from pandas.api.types import pandas_dtype
+from pandas.core.arrays import ArrowExtensionArray, ExtensionArray  # type: ignore[attr-defined]
+from pandas.core.indexers import (  # type: ignore[attr-defined]
+    check_array_indexer,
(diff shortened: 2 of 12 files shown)


src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+    with no alignment padding or per-tensor switching to scientific notation.
+    """
+    if _is_na(value):
+        return str(pd.NA)
+    if value.size > TENSOR_FORMATTING_MAX_ELEMENTS:
+        return _tensor_descriptor(value)
+    return str(value.tolist())
+
+
+def _is_na(value: Any) -> bool:
```
hombit, line 88: Can we use `pd.isna()` instead?

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+    raise TypeError(
+        "Expected TensorDtype, pa.FixedShapeTensorType, pd.ArrowDtype of a tensor type or a tensor dtype "
+        f"string, got {dtype!r}"
+    )
+
+
+def _storage_of(array: pa.ChunkedArray) -> pa.ChunkedArray:
+    """Fixed-size-list storage of a tensor-typed chunked array."""
+    pa_type = cast(pa.FixedShapeTensorType, array.type)
+    return pa.chunked_array([chunk.storage for chunk in array.iterchunks()], type=pa_type.storage_type)
```
hombit: I guess it may be inlined with `array.cast(array.type.storage_type)`

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+
+
+def _is_tensor_scalar(value: Any) -> bool:
+    """Whether a value is a pyarrow scalar of a fixed_shape_tensor type.
+
+    We check the type rather than ``isinstance(value, pa.FixedShapeTensorScalar)``
+    because that class is not exported by pyarrow before version 17.
```
hombit: Let's just bump to 17. Anyway we will probably bump to 26 soon...

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+    return isinstance(value, pa.ExtensionScalar) and isinstance(value.type, pa.FixedShapeTensorType)
+
+
+def _tensor_scalar_to_numpy(scalar: pa.ExtensionScalar) -> np.ndarray:
+    """Read-only zero-copy numpy view of a valid fixed_shape_tensor scalar.
+
+    ``scalar.value`` is the fixed_size_list storage scalar, whose ``values``
+    are a slice of the flat child array. This is what
+    ``FixedShapeTensorScalar.to_numpy()`` does, but that method only exists
+    since pyarrow 17.
```
hombit: Same, let's bump pyarrow

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+    """
+    return np.asarray(scalar.value.values).reshape(tuple(scalar.type.shape))
+
+
+def _tensor_to_flat(value: Any, dtype: TensorDtype) -> np.ndarray:
+    """Validate a single tensor against the dtype and return its values flattened in C order."""
+    if _is_tensor_scalar(value):
+        array = _tensor_scalar_to_numpy(value)
+    elif isinstance(value, pa.Scalar):
+        array = np.asarray(value.as_py())
```
hombit: I think we should try to view data here, e.g. the following copies for Null values only:
```suggestion
        array = np.asarray(value.value.values)
```

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+        # A fixed_size_list scalar comes back flat
+        if array.shape != dtype.shape and array.size == dtype.size:
```
hombit: I was a bit confused about the flow here. 
- How could sizes mismatch? Should we raise an error if they do?
- Should we reshape unconditionally
- Why do we reshape it here, and then reshape again bellow?

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+            array = array.reshape(dtype.shape)
+    else:
+        array = np.asarray(value)
+    if array.shape != dtype.shape:
+        raise ValueError(f"Expected a tensor of shape {dtype.shape}, got shape {array.shape}")
```
hombit: Should we reshape this one?

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+    :meth:`to_stack` return read-only numpy views over the arrow buffers
+    whenever possible, so no data is copied.
+
+    Parameters
+    ----------
+    values : pyarrow.Array or pyarrow.ChunkedArray
+        The array to be wrapped. Either a ``fixed_shape_tensor`` extension
+        array, or its ``fixed_size_list`` storage array, in which case
+        ``dtype`` is required.
+    dtype : TensorDtype, optional
```
hombit, line 203: Maybe I'm overthinking, but it may be useful to either raise when `dtype is not None` and `values` is a tensor array, or allow to cast it to given `dtype`.

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+        else:
+            raise ValueError(
+                f"values must be a fixed_shape_tensor or fixed_size_list array, got type {values.type}"
+            )
+
+        self._pa_array = values
+        self._dtype = dtype
+
+    @staticmethod
+    def _wrap_storage(storage: pa.ChunkedArray, dtype: TensorDtype) -> pa.ChunkedArray:
```
hombit: Move out of class?

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+            return cls(scalars, dtype=dtype)
+        if isinstance(scalars, np.ndarray) and scalars.dtype != np.object_ and scalars.ndim >= 2:
+            if dtype is not None and scalars.ndim == dtype.ndim:
+                # A single tensor (possibly of the wrong shape, which is reported below), e.g. pandas
+                # broadcasting a "scalar", rather than a stack of them
+                scalars = [scalars]
+            else:
+                return cls.from_stack(scalars, dtype=dtype)
+
+        scalars = list(scalars)
```
hombit, line 295: Would we go through the list on the assignment time? That should be pretty slow...

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+
+        Two tensors are equal when they have the same dtype and all their values
+        are equal, with numpy's semantics for the values (so NaN != NaN). The
+        result is NA where either side is missing. Comparing to anything that
+        cannot be interpreted as tensors of this dtype, including an array of a
+        different tensor dtype, gives False everywhere.
+
+        Returns
+        -------
+        pd.arrays.BooleanArray
```
hombit: Let's return `ArrowExtensionArray` of bool dtype if we strongly decide to do row-wise comparison

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+        self._pa_array = self._wrap_storage(storage, self.dtype)
+
+    def __len__(self) -> int:
+        return len(self._pa_array)
+
+    def __iter__(self) -> Iterator[np.ndarray | Any]:
+        for scalar in self._pa_array:
+            yield self._scalar_to_numpy(scalar)
+
+    def __eq__(self, other):  # type: ignore[override]
```
hombit, line 441: This has too many edgecases, including null tensor elements. Let's raise NonImplemented or apply the opearation element-wise and return `Self` of bool dtype

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+
+def _tensor_scalar_to_numpy(scalar: pa.ExtensionScalar) -> np.ndarray:
+    """Read-only zero-copy numpy view of a valid fixed_shape_tensor scalar.
+
+    ``scalar.value`` is the fixed_size_list storage scalar, whose ``values``
+    are a slice of the flat child array. This is what
+    ``FixedShapeTensorScalar.to_numpy()`` does, but that method only exists
+    since pyarrow 17.
+    """
+    return np.asarray(scalar.value.values).reshape(tuple(scalar.type.shape))
```
hombit, line 163: Partially overlaps with `_tensor_to_flat`

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+        return self._wrap_result(cast(pa.ChunkedArray, scalar_or_array))
+
+    def _wrap_result(self, pa_array: pa.ChunkedArray) -> Self:  # type: ignore[name-defined] # noqa: F821
+        """Wrap a selection from this array, carrying over the dtype and the read-only flag."""
+        result = type(self)(pa_array, dtype=self.dtype)
+        result._readonly = self._readonly
+        return result
+
+    def __setitem__(self, key, value) -> None:
+        if self._readonly:
```
hombit, line 370: It is not consistent with the NestedSeries, where we basically re-create the full series to set an element. We probably should support the same thing here, or at least tell the user how to make it read-write and open an issue to support it later.

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+            # Don't modify in-place
+            indices_array = np.copy(indices_array)
+            indices_array[indices_array < 0] += len(self)
+        return type(self)(self._pa_array.take(pa.array(indices_array)), dtype=self.dtype)
+
+    def copy(self) -> Self:  # type: ignore[name-defined] # noqa: F821
+        """Return a copy of the extension array.
+
+        This implementation returns a shallow copy of the extension array,
+        because the underlying PyArrow array is immutable.
```
hombit, line 653: I guess we were wrong about it in the Nested array implementation, and in other places of the code you check if the data readonly

src/nested_pandas/tensors/ext_array.py
```diff
@@ -0,0 +1,900 @@
...
+        ----------
+        na_value : Any, default np.nan
+            Value to fill missing tensors with.
+
+        Returns
+        -------
+        np.ndarray
+            Array of shape ``(len(self), *self.dtype.shape)``.
+        """
+        stack = self._values_block()
```
hombit, line 880: Here we may cast integers to floats

hombit (commented):
Huge and nice job, @smcguire-cmu!

> Some methods in nested were overwriting superclass methods just to call the method and raise errors for not being implemented. I dropped those since they're inherited anyway: they only called super(), which raises AbstractMethodError, so behaviour is identical with or without them. If they have a reason to be there I can add them back.

I put them there to explicitly implement all the methods, so when the upstream base class is updated, we can see which methods were added and don't have to guess what and why is missing. I'd keep all of them for this reason.

I'd refuse any conversion of integer tensors with nulls to numpy when `na_value == np.nan`, since it 1) implicitly changes the dtype to float, 2) and thus corrupts 64-bit integer data. We should change the code to support proper casting in the case of integer dtype and integer `na_value`. That would mean that we should do something like this in all numpy-casting calls (like scalar casting, `_values_block`, etc):
```python
flat = something.values.slice(...)
<... check type of the fill_value vs type of the array ...>
if flat.null_count > 0:
    flat = pa.compute.fill_na(flat, fill_value)
# should never fail, because there are no nulls anymore
numpy_array = flat.to_numpy(zero_copy_only=False)
```

src/nested_pandas/tensors/ext_array.py
smcguire-cmu: Yeah, I think it's complicated, we can come to it during the math operations issue for now just not implemented

src/nested_pandas/tensors/ext_array.py
smcguire-cmu: In the implementation, yes, but this method needs some extra guards for a) pa types, and b) making numpy arrays not return a bool array, which actually weren't quite correct before anyway

src/nested_pandas/tensors/ext_array.py
smcguire-cmu: Partially, but tensor_to_flat also has the validation and needs the dtype since it needs to work with different input types, so I don't think it's worth modifying it to save the `np.asarray(scalar.value.values)` repeating.

src/nested_pandas/tensors/ext_array.py
smcguire-cmu: Yeah this method is kinda sloppy tbh. The reshape here and below is because here it reshapes so the dtype shape check can run, then it reshapes to flat, but I agree I don't like this behavior.

I rewrote it to be more explicit about its handling of different input types, and just ignore reshaping and raise an error if it's not the same shape as the dtype. I think that makes more sense to me than silently reshaping if the user e.g. makes an error and uses the wrong dtype but their 2x3 gets reshaped into a 3x2.

src/nested_pandas/tensors/ext_array.py
smcguire-cmu: Yeah, we currently cast, but edge cases with permutations are weird, let's just error

src/nested_pandas/tensors/ext_array.py
smcguire-cmu: This is a pandas (I think introduced in pandas 3) property on the extension array, and this matches pandas expected behavior for it. TensorExtensionArray implements the pandas test suite for extension arrays and this passes their checks, so I think it's good to leave it like this and nested should be updated with the same behavior. I don't think pandas actually uses the property anywhere yet and it's just intended for future changes, but it is tested so it's worth having I think.

src/nested_pandas/tensors/ext_array.py
smcguire-cmu: In the other places where we're checking readonly it's not that we're checking if the pyarrow data is read only, but that if the pandas attribute wants our array to act as readonly. I think pyarrow arrays should always be immutable unless their buffers refer to data outside of their control e.g. zero copy from numpy. At least pandas ArrowExtensionArray behaves this way thinking the arrays are immutable.

src/nested_pandas/tensors/ext_array.py
smcguire-cmu: I made an issue to do the test suites for nested, I can make a separate one for the readonly property if you want.

smcguire-cmu:
I've added the methods back, and we're now rejecting at construction time any arrays with inner nulls that aren't floats. I've had to bump the minimum pyarrow version up to 18 for the FixedShapeTensor support, which also means the comment there ` # remove struct_field_names() and struct_fields() when upgraded to 18+` Now applies

pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
     "numpy>=2",
     # We use internal pd._libs.missing and experimental ArrowExtensionArray
     "pandas>=3,<4",
-    "pyarrow>=16",  # remove struct_field_names() and struct_fields() when upgraded to 18+
+    "pyarrow>=18",  # remove struct_field_names() and struct_fields() when upgraded to 18+
```
hombit, line 24: Please remove these helpers or open an issue

hombit (approved):
Thank you!

